PR #2, opened 2020-06-29
Add beetle-step and React Components

wdcroslen:


jiaxinz-g (commented):
We are not going to use all of these components. Let's just add the ones that are definitely going to be used as a start, and add more later on when needed. I am afraid these 33k files are hard to manage and migrate.

sasymer:
> We are not going to use all of these components. Let's just add the ones that are definitely going to be used as a start, and add more later on when needed. I am afraid these 33k files are hard to manage and migrate.

I agree, and I think a lot of those files are made when you make a React app. How do we know which ones to keep?

PR #3, opened 2020-06-30
Delete target folder files

sasymer:
- From last merge/my last commits, added target folder files. We don't need these files in the repo, so this pull request is to delete them.

@jiaxinz-g and @gagomez could you please confirm this is the right thing to do?


jiaxinz-g:
I prefer not having target folder in the repo. Thank you for the change, Sanna!

jiaxinz-g (approved):

jiaxinz-g (approved):

wdcroslen (approved):

PR #4, opened 2020-07-01
Create dropdown that will list databases.

### LINKED ISSUE #7: Database selection
Load databases from the servlet, create a selector for databases. Users can only select a single database.

hilakey:
Here's how the database list will look on a live server: [](url) https://play-user-data-beetle.uc.r.appspot.com/
+ list of databases has been temporarily hard coded
+ Modify formatting
+Remove white spaces
+Rename files


src/main/java/com/google/sps/servlets/exampledbServlet.java
```diff
@@ -0,0 +1,77 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+/** Servlet that returns HTML that contains content from the example database. */
+@WebServlet("/example")
+public class exampledbServlet extends HttpServlet {
```
gagomez: Upper camelcase here and in the file name.

Also, is this an example or is this the actual get databases servlet?

src/main/java/com/google/sps/servlets/exampledbServlet.java
```diff
@@ -0,0 +1,77 @@
...
+import com.google.gson.Gson;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+/** Servlet that returns HTML that contains content from the example database. */
```
gagomez: This doesn't return HTML right? It returns the list of databases configured for beetle.

src/main/java/com/google/sps/servlets/exampledbServlet.java
```diff
@@ -0,0 +1,77 @@
...
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    response.setContentType("application/JSON;");
+
+    List<String> listDatabases = new ArrayList<String>();
+
+    try (ResultSet resultSet =
+        dbClient
+            .singleUse() // Execute a single read or query against Cloud Spanner.
+            .executeQuery(Statement.of("SELECT option_name FROM information_schema.database_options WHERE catalog_name = '' and schema_name = ''"))) {
```
gagomez: These are options set on a single database like permissions and such. I think what we want here is a list of databases right? We won't get that by querying a single database. For now, let's just have this be a hardcoded list of IDs, later we can look into using a configuration file to specify these.

src/main/webapp/index.html
```diff
@@ -3,9 +3,18 @@
...
+  <body onload="getDatabases()">
     <h1>Play User Data Beetle</h1>
     <p>Click <a href="/example-db">here</a> to view data queried from an example Cloud Spanner database.</p>
+    <p>please select the database you would like to use</p>
+
+    <div id="select-db">
+        <select id="list-databases" value="list-databases" class="select">
+            <option id="select-title">Select a Database</option>
+        </select>
+    </div>
```
gagomez: Should we start with a "Loading..." message until the list is retrieved?

src/main/webapp/index.html
```diff
@@ -3,9 +3,18 @@
...
+    <script src="script.js"></script>
   </head>
-  <body>
+  <body onload="getDatabases()">
     <h1>Play User Data Beetle</h1>
     <p>Click <a href="/example-db">here</a> to view data queried from an example Cloud Spanner database.</p>
+    <p>please select the database you would like to use</p>
+
+    <div id="select-db">
+        <select id="list-databases" value="list-databases" class="select">
```
gagomez: What does setting the value to list-databases do? It isn't even an option in the select.

src/main/java/com/google/sps/servlets/exampledbServlet.java
```diff
@@ -0,0 +1,77 @@
...
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+/** Servlet that returns HTML that contains content from the example database. */
+@WebServlet("/example")
```
gagomez: Let's come up with a more descriptive path. For example, "/databases"

src/main/webapp/script.js
```diff
@@ -0,0 +1,37 @@
+
+function getDatabases(){
+ 
+    fetch("/example").then(response => response.json()).then((list) => {
+      console.log(list);
+
+      let dropdown = document.getElementById('list-databases');
+      dropdown.length = 0;
```
gagomez: This seems odd. I am guessing you are doing this to clear the options set in the dropdown?

I would suggest `dropdown.options.remove(0)` as it is considerably easier to read.

src/main/webapp/script.js
```diff
@@ -0,0 +1,37 @@
...
+            if(list[i] != null){
+                dropdown.add(option);
+                console.log(list[i] + "was added into the dropdown");
+            }else{
+                console.log(list[i] + "was  not added into the dropdown");
+            }
+      	  
+        }
+
+
```
gagomez: Let's remove this unnecessary whitespace as well as all these console logs which I imagine where for dev testing.

gagomez (commented):
Remove all the additions from target/* as those are built from the source which is what we want to have in github.

src/main/java/com/google/sps/servlets/exampledbServlet.java
```diff
@@ -0,0 +1,77 @@
...
+        dbClient
+            .singleUse() // Execute a single read or query against Cloud Spanner.
+            .executeQuery(Statement.of("SELECT option_name FROM information_schema.database_options WHERE catalog_name = '' and schema_name = ''"))) {
+       while(resultSet.next()){
+           /*response
+            .getWriter()
+            .println(
+                String.format(
+                    "<p>%s</p>",
+                    resultSet.isNull(0) ? "" : resultSet.getString(0)));*/
```
sasymer: Maybe could get rid of this comment unless you're going to use it?

src/main/webapp/index.html
sasymer: I like that idea too!

target/beetle-1/index.html
```diff
@@ -0,0 +1,20 @@
...
+  <body onload="getDatabases()">
+    <h1>Play User Data Beetle</h1>
+    <p>Click <a href="/example-db">here</a> to view data queried from an example Cloud Spanner database.</p>
+    <p>please select the database you would like to use</p>
+
+    <div id="select-db">
+        <select id="list-databases" value="list-databases" class="select">
+            <option id="select-title">Select a Database</option>
+        </select>
+    </div>
```
sasymer: Indentation

src/main/java/com/google/sps/servlets/exampledbServlet.java
```diff
@@ -0,0 +1,77 @@
...
+/** Servlet that returns HTML that contains content from the example database. */
+@WebServlet("/example")
+public class exampledbServlet extends HttpServlet {
+
+  DatabaseClient dbClient;
+
+  public void init() {
+    Spanner spanner = SpannerOptions.newBuilder().build().getService();
+    DatabaseId db = DatabaseId.of("play-user-data-beetle", "test-instance" , "example-db");
+    this.dbClient = spanner.getDatabaseClient(db);
```
sasymer: Eventually (when hard coding is taken out), probably won't need this part because I don't think you need to open specific databases. I'm not sure about this, though, because you use the dbClient below.

src/main/java/com/google/sps/servlets/exampledbServlet.java
```diff
@@ -0,0 +1,77 @@
...
+        String json = convertToJsonUsingGson(listDatabases);
+        response.getWriter().println(json);
+       }
+  }
+
+  private String convertToJsonUsingGson(List<String> listDatabases) {
+    Gson gson = new Gson();
+    String json = gson.toJson(listDatabases);
+    return json;
+  }
```
sasymer: Nice helper method

sasymer (changes requested):
I think target files should be deleted before merging (per what Jiaxin said on my pull request)

src/main/java/com/google/sps/servlets/exampledbServlet.java
hilakey: This is the actual get servlet, but I just didn't give it a more conventional name at the time. But I just fixed it.

src/main/java/com/google/sps/servlets/exampledbServlet.java
hilakey: Originally it was HTML but it actually returns JSON data, so I updated the comment to reflect this!

src/main/java/com/google/sps/servlets/exampledbServlet.java
hilakey: All right, done. As I was looking for documentation on listing databases it seems cloud spanner has...very little information on this.

src/main/webapp/index.html
hilakey: Done!

src/main/webapp/index.html
hilakey: That is true, I guess I'm used to examples having values for many elements even if they aren't necessary, but I removed the value.

src/main/java/com/google/sps/servlets/exampledbServlet.java
hilakey: Done

src/main/webapp/script.js
hilakey: Yes that's what it does, I wasn't familiar with this other option you showed but I do see how it's easier to read! Thanks for the tip just changed it!

src/main/webapp/script.js
hilakey: Done

src/main/java/com/google/sps/servlets/exampledbServlet.java
hilakey: Done

sasymer (approved):

wdcroslen (approved):
Looks good! I like the logic in your Javascript and how the databases are added to the dropdown!

src/main/java/com/google/sps/servlets/BeetleDbServlet.java
```diff
@@ -0,0 +1,64 @@
...
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+/** Servlet that returns JSON that contains a list of databases. */
+@WebServlet("/databases")
+public class BeetleDbServlet extends HttpServlet {
+
+  DatabaseClient dbClient;
```
jiaxinz-g, line 36: Since we are hardcoding the list of DBs to return, these are not needed. Let's remove them, so as the corresponding unused imports.

src/main/webapp/Filters.html
```diff
@@ -0,0 +1,6 @@
+<!DOCTYPE html>
```
jiaxinz-g: Let's not add this file until it is being used, would be better to keep it in the local change if you plan to use it in a followup PR.

jiaxinz-g (approved):
Two minor comments. The rest looks good to me!

src/main/webapp/Filters.html
hilakey: Ah okay just un-added it.

src/main/java/com/google/sps/servlets/BeetleDbServlet.java
hilakey: Done.

gagomez (approved):

PR #5, opened 2020-07-02
Finish table selection dropdown. Able to print data from database onto page.

### LINKED ISSUE #8: Table Selection
Load tables from the selected database, and create a selector for tables. Users can select multiple tables.

sasymer:
- Create servlet for table selection
- Create servlet for data displaying on page, given a list of tables to query from (this is given by the table selection servlet)
- Hopefully this works - trying to do this request from a new branch


src/main/java/com/google/sps/servlets/DataFromDatabase.java
```diff
@@ -0,0 +1,134 @@
...
+
+@WebServlet("/data-from-db")
+public class DataFromDatabase extends HttpServlet {
+
+  DatabaseClient dbClient;
+  private String[] selectedTables;
+
+  public void init() {
+    Spanner spanner = SpannerOptions.newBuilder().build().getService();
+    DatabaseId db = DatabaseId.of("play-user-data-beetle", "test-instance", "example-db");
```
gagomez: Since we need to support multiple databases, is this something that should be read from the request?

src/main/java/com/google/sps/servlets/DataFromDatabase.java
```diff
@@ -0,0 +1,134 @@
...
+
+    for (String table : selectedTables) {
+      String colQuery = "SELECT column_name, spanner_type, is_nullable FROM information_schema.columns WHERE table_name = '" + table + "'";
+
+      try (ResultSet resultSet =
+        dbClient.singleUse().executeQuery(Statement.of(colQuery))) {
+
+        List<String> colnames = new ArrayList<>();
+        List<String> spannerTypes = new ArrayList<>();
+        List<String> nullables = new ArrayList<>();
```
gagomez: Would it be clearer if we had an object representation of "ColumnMetadata"? If so, take a look at AutoValue it may save you some time: https://github.com/google/auto/blob/master/value/userguide/index.md

src/main/java/com/google/sps/servlets/DataFromDatabase.java
```diff
@@ -0,0 +1,134 @@
...
+    Spanner spanner = SpannerOptions.newBuilder().build().getService();
+    DatabaseId db = DatabaseId.of("play-user-data-beetle", "test-instance", "example-db");
+    this.dbClient = spanner.getDatabaseClient(db);
+  }
+
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    response.setContentType("text/html;");
+
+    for (String table : selectedTables) {
```
gagomez: How does selectedTables get populated?

src/main/java/com/google/sps/servlets/DataFromDatabase.java
```diff
@@ -0,0 +1,134 @@
...
+      String colQuery = "SELECT column_name, spanner_type, is_nullable FROM information_schema.columns WHERE table_name = '" + table + "'";
+
+      try (ResultSet resultSet =
+        dbClient.singleUse().executeQuery(Statement.of(colQuery))) {
+
+        List<String> colnames = new ArrayList<>();
+        List<String> spannerTypes = new ArrayList<>();
+        List<String> nullables = new ArrayList<>();
+
+        while (resultSet.next()) {
```
gagomez: Should we return an error if there are no columns? (e.g. table doesn't exist in database)

src/main/java/com/google/sps/servlets/DataFromDatabase.java
```diff
@@ -0,0 +1,134 @@
...
+  private void executeTableQuery(String query, List<String> colnames, List<String> spannerTypes, HttpServletResponse response) throws IOException {
+    try (ResultSet rs =
+      dbClient
+      .singleUse() // Execute a single read or query against Cloud Spanner.
+      .executeQuery(Statement.of(query))) {
+
+      while (rs.next()) {
+        String row = "";
+        int index = 0;
+        while (index < colnames.size()) {
```
gagomez: nit: this seems like an appropriate time to use a for loop

src/main/java/com/google/sps/servlets/DataFromDatabase.java
```diff
@@ -0,0 +1,134 @@
...
+            case "BYTES(MAX)":
+              ByteArray bytes = rs.getBytes(colName);
+              byte[] byteArray = bytes.toByteArray();
+          
+              for (byte b : byteArray) {
+                row += b;
+              }
+              break;
+            case "BYTES":
+              row += rs.getBytes(colName);
```
gagomez: Here and above let's make sure that we are encoding the bytes to UTF8 strings

src/main/java/com/google/sps/servlets/DataFromDatabase.java
```diff
@@ -0,0 +1,134 @@
...
+        }
+        response.getWriter().println(row); 
+      }
+    }
+  }
+
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    selectedTables = request.getParameterValues("table-select");
+    response.sendRedirect("/select-tables.html");
```
gagomez: Instead of doing a POST to set selectedTables and then a GET to read from them can we not pass the selected tables in the GET request?

NOTE: this will eventually run on a number of machines where requests could go to any machine. You are keeping server side state here and when it runs on multiple machines this will break since the second request could go to a whole different machine (or worse, someone else's post could have changed selectedTables).

src/main/java/com/google/sps/servlets/TablesFromDatabase.java
```diff
@@ -0,0 +1,56 @@
...
+/** Servlet that returns HTML that contains tables in the example database. */
+@WebServlet("/tables-from-db")
+public class TablesFromDatabase extends HttpServlet {
+
+  DatabaseClient dbClient;
+
+  public void init() {
+    Spanner spanner = SpannerOptions.newBuilder().build().getService();
+
+    //TODO: change "example-db" to the database that Millennia passes in from the db selection
```
gagomez: Note, we can do more here by hardcoding this in the javascript rather than the server. That way we don't need to change the server at all.

gagomez (commented):
Let's also ignore the files in the target directory

src/main/java/com/google/sps/servlets/DataFromDatabase.java
sasymer: Yes! I had it this way (hardcoded) for testing my feature before I had access to the database info in the request.

src/main/java/com/google/sps/servlets/DataFromDatabase.java
sasymer: From a post request (in the doPost below). When the form is submitted a post request happens, sending the selected values to the java servlet. I tried putting it in the get request instead, but the values were always null. Is there a better way around this?

sasymer: Fixed this - now gets populated in the doGet

src/main/java/com/google/sps/servlets/DataFromDatabase.java
sasymer: Done

src/main/java/com/google/sps/servlets/DataFromDatabase.java
sasymer: Done

src/main/java/com/google/sps/servlets/DataFromDatabase.java
sasymer: Done

src/main/java/com/google/sps/servlets/TablesFromDatabase.java
sasymer: Got rid of init() method

src/main/java/com/google/sps/servlets/DataFromDatabase.java
sasymer: I think so! I looked into this today and will work on implementing it tomorrow.

src/main/java/com/google/sps/servlets/DataFromDatabase.java
```diff
@@ -0,0 +1,159 @@
...
+import java.util.List;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+@WebServlet("/data-from-db")
+public class DataFromDatabase extends HttpServlet {
+
+  private static final String TABLE_SELECT_PARAM = "table-select";
```
wdcroslen: I like the use of these variables to make the code simpler!

src/main/webapp/database-tables.js
```diff
@@ -0,0 +1,75 @@
+function showDatabase() {
+  const search = window.location.search;
+  const queryString = '/data-from-db' + search;
+  console.log(queryString)
```
wdcroslen: Maybe the query string is something we don't need to log because we can see the URL?

wdcroslen (approved):
The changes overall look really nice!

src/main/java/com/google/sps/servlets/DataFromDatabase.java
sasymer: Thanks!

src/main/webapp/database-tables.js
sasymer: I think I did that because at this point the page we're on is main-page.html. So I get the part after the ? and concat it to the data-from-db

src/main/java/com/google/sps/servlets/DataFromDatabase.java
sasymer: Doing this in separate PR

src/main/java/com/google/sps/servlets/DataFromDatabase.java
sasymer: I'll add this in the next PR if that's ok? I modified this part in a new branch to use the AutoValue object, and I think if I do this error checking there it'll reduce merge conflicts.

src/main/java/com/google/sps/servlets/DataFromDatabase.java
sasymer: Done.

src/main/java/com/google/sps/servlets/DataFromDatabase.java
```diff
@@ -0,0 +1,175 @@
...
+      try (ResultSet resultSet =
+        dbClient.singleUse().executeQuery(Statement.of(columnQuery))) {
+        List<Schema> schemas = new ArrayList<>();
+ 
+        while (resultSet.next()) {
+          schemas.add(createSchema(resultSet));
+        }
+ 
+        // No columns -> throw error
+        if (schemas.size() == 0) {
```
hilakey: I like the runtime error!

hilakey (approved):
Looking good! The changes are really nice, could you add a link to the live server to see the implementation :)

sasymer:
http://play-user-data-beetle.appspot.com/

src/main/java/com/google/sps/servlets/DataFromDatabase.java
```diff
@@ -0,0 +1,175 @@
...
+import java.util.List;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+@WebServlet("/data-from-db")
+public class DataFromDatabase extends HttpServlet {
+
+  private static final String COMMA = ", ";
```
jiaxinz-g: Great idea of creating these static final variables for queries! Consider creating a package level Constants.java file that holds these variables so that they can be shared within the project. Maybe the list of databases ["example-db", "example-db-2", "example-db-3", "example-db-4"] can go there too (if not another package level shared file).

src/main/java/com/google/sps/servlets/DataFromDatabase.java
```diff
@@ -0,0 +1,175 @@
...
+  private static final String TABLE_SELECT_PARAM = "table-select";
+  private static final String TEST_INSTANCE = "test-instance";
+  private static final String TEXT_TYPE = "text/html;";
+
+  DatabaseClient dbClient;
+  private String[] selectedTables;
+
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    selectedTables = request.getParameterValues(TABLE_SELECT_PARAM);
```
jiaxinz-g: Connecting to databases are expensive operations, we can also feel the delay in fetching the list of tables. Could we create a file that is called DatabaseConnector.java, that initializes all the database clients, and provides APIs to return the database client given a database name? Since our list of databases are hardcoded for now, I think it would be fine if we initialize them all at once at server starts.

src/main/java/com/google/sps/servlets/DataFromDatabase.java
```diff
@@ -0,0 +1,175 @@
...
+import java.io.UnsupportedEncodingException;
+import java.util.ArrayList;
+import java.util.List;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+@WebServlet("/data-from-db")
+public class DataFromDatabase extends HttpServlet {
```
jiaxinz-g: Add Servlet suffix to this class name. Now we have many files under the same servlets folder, it would be easier to identify which ones are servlets that handles the requests, which ones are helper classes.

src/main/java/com/google/sps/servlets/DataFromDatabase.java
```diff
@@ -0,0 +1,175 @@
...
+    return query;
+  }
+
+  private void initializeDatabase(String databaseName) {
+    Spanner spanner = SpannerOptions.newBuilder().build().getService();
+    DatabaseId db = DatabaseId.of(PROJECT, TEST_INSTANCE, databaseName);
+    this.dbClient = spanner.getDatabaseClient(db);
+  }
+
+   private void executeTableQuery(Table tableObject, String query, List<Schema> schemas, HttpServletResponse response) throws IOException {
```
jiaxinz-g: This HttpServletResponse response parameter seems not being used in the method.

src/main/java/com/google/sps/servlets/DataFromDatabase.java
```diff
@@ -0,0 +1,175 @@
...
+  }
+
+  private Schema createSchema(ResultSet resultSet) {
+    String columnName = resultSet.getString(0);
+    String schemaType = resultSet.getString(1);
+    String nullable = resultSet.getString(2);
+    return Schema.create(columnName, schemaType, nullable);
+  }
+
+  private String constructQueryString(List<Schema> schemas, Table tableObject, String table) {
```
jiaxinz-g: We tend to avoid concatenating strings in Java, it is okay when there are only a few Strings, but when there is a for loop, using StringBuilder to build the concatenated string is more efficient and a better choice.  

Here we are building the raw String used in the query, and then pass it in Statement.of(). Instead, we can use the Cloud Spanner Statement and Statement.Builder to build the query to execute. 

https://googleapis.dev/java/google-cloud-spanner/latest/com/google/cloud/spanner/Statement.html
https://googleapis.dev/java/google-cloud-spanner/latest/com/google/cloud/spanner/Statement.Builder.html

You could create a Statement.Builder statementBuilder variable, and return statementBuilder.build() in this method.

src/main/java/com/google/sps/servlets/DataFromDatabase.java
```diff
@@ -0,0 +1,175 @@
...
+        tableObject.addRow(rowObject);
+      }
+    }
+  }
+
+  private void addDataToRowObject(String dataType, Row rowObject, String columnName, ResultSet resultSet) {
+    switch (dataType) {
+      case "STRING(MAX)":
+      case "STRING(250)":
+      case "STRING(1024)":
```
jiaxinz-g: This needs to be updated when we have a new column that is defined with a different length, and that would be a pain to maintain. Instead of doing the exact match on the dataType String, probably it is better to extract the type without length, e.g "STRING", "BYTES" before the switch. 

Another option if you want to go is to use the Enum Type.Code https://googleapis.dev/java/google-cloud-spanner/latest/com/google/cloud/spanner/Type.Code.html. I don't find a good way to get the Type code enum from the String, maybe you will need to create a helper method yourself, something like getTypeCodeEnumFromString, and return the enum, and use it here int he switch. You might need extra logic handling when the type code is ARRAY and check the inner type of the array. The benefit of using this enum is that we won't worry about missing the logic to handle a type. 

I am totally fine with the just using the dataType without the length .

src/main/java/com/google/sps/servlets/DataFromDatabase.java
```diff
@@ -0,0 +1,175 @@
...
+      case "ARRAY<INT64>":
+        String arrayToString = longArrayToString(resultSet.getLongArray(columnName));
+        rowObject.addData(columnName, arrayToString);
+        break;
+      case "DATE":
+        rowObject.addData(columnName, EMPTY_STRING + resultSet.getDate(columnName));
+        break;
+      case "BOOL":
+        rowObject.addData(columnName, EMPTY_STRING + resultSet.getBoolean(columnName));
+    }
```
jiaxinz-g: Let's have a default case, where we just add a string saying this type is not supported yet, or some other way to get informed when we need to update this code here. 

Also please consider ordering the cases in a way that all primitive types first, and then TIMESTAMP, DATE, ARRAY. 

We probably need to come back and update the part that handing the array for other types.

src/main/java/com/google/sps/servlets/DataFromDatabase.java
```diff
@@ -0,0 +1,175 @@
...
+  }
+
+  private String longArrayToString(long[] longArray) {
+    String arrayToString = "[";
```
jiaxinz-g: Consider using StringBuilder.

src/main/java/com/google/sps/servlets/Table.java
```diff
@@ -0,0 +1,26 @@
+package com.google.sps.servlets;
+ 
+import java.util.ArrayList;
+import java.util.List;
+
+public class Table {
+  private String name;
+  private List<String> columns = new ArrayList<String>();
+  private List<Row> rows = new ArrayList<Row>();
```
jiaxinz-g, line 10: I can see how Row made of a map helps with the table structure. But in reality this is expensive, and we can get the column name to the value string by mapping the index. For example, if the Table has

List<String> columns
List<List<String>> content

This will have enough information to display and not hard to iterate. e.g. In database-tables.js line 47, 48, you can get dataPoint with colIndex. 

I think this will simplify our code and make it more efficient, and if in the future, we do need Row and a map in the row, we can refactor it later.

src/main/java/com/google/sps/servlets/Table.java
```diff
@@ -0,0 +1,26 @@
+package com.google.sps.servlets;
+ 
+import java.util.ArrayList;
+import java.util.List;
+
+public class Table {
```
jiaxinz-g, line 7: Table can also be an Autovalue, and you probably need it with a builder . https://github.com/google/auto/blob/master/value/userguide/builders.md

jiaxinz-g (commented):
Sending out comments I just have, will finish the review tomorrow.

src/main/java/com/google/sps/servlets/Table.java
jiaxinz-g: It is totally fine if you keep the Row class, and just change Map<String, String> row to List<String> values. But please consider also creating Row as an autovalue.

src/main/java/com/google/sps/servlets/DataFromDatabase.java
sasymer: Done - threw a runtime exception if no columns for now

src/main/java/com/google/sps/servlets/DataFromDatabase.java
sasymer: Should Constants.java be a static class? I will put additional ideas for this in the design ideas doc.

src/main/java/com/google/sps/servlets/DataFromDatabase.java
sasymer: I like this idea a lot! I'll write out a design plan for it.

src/main/java/com/google/sps/servlets/DataFromDatabase.java
sasymer: Should we put helper classes in a separate Java package?

sasymer: Done

src/main/java/com/google/sps/servlets/DataFromDatabase.java
sasymer: Fixed, thank you

src/main/java/com/google/sps/servlets/Table.java
sasymer: Ok that makes sense. Just to clarify, does having a map in the Row object make it more expensive in terms of memory or time?

src/main/java/com/google/sps/servlets/DataFromDatabase.java
sasymer: Ok I see-- that makes a lot of sense in terms of flexibility and extendibility for future additions.

sasymer: Does this apply also to INT64 or should that stay the same?

src/main/java/com/google/sps/servlets/DataFromDatabase.java
jiaxinz-g: This Constants.java class is just a regular class, but its variables are all static.

src/main/java/com/google/sps/servlets/DataFromDatabase.java
sasymer: Done (all except for updating array handling)

src/main/java/com/google/sps/servlets/DataFromDatabase.java
sasymer: Done (for string and bytes)

src/main/java/com/google/sps/servlets/DataFromDatabase.java
jiaxinz-g: We can leave them in the same package.

src/main/java/com/google/sps/servlets/DataFromDatabase.java
jiaxinz-g: INT64 is a type itself, so it can stay the same.

src/main/java/com/google/sps/servlets/Table.java
jiaxinz-g: Yeah, Maps are more expensive, mostly in terms of memory.

src/main/java/com/google/sps/servlets/DataFromDatabase.java
sasymer: Ok, done! Made a class Constants.java that just has default static variables and its used in the Data Servlet and Tables Servlet.

src/main/java/com/google/sps/servlets/DataFromDatabase.java
sasymer: Done in new PR

src/main/java/com/google/sps/servlets/DataFromDatabase.java
sasymer: Added this to my new PR.

src/main/java/com/google/sps/servlets/Constants.java
```diff
@@ -0,0 +1,20 @@
+package com.google.sps.servlets;
+
+public class Constants {
```
jiaxinz-g: We could also make this class package-private by removing public keyword. It is recommended to start with more restricted visibilities, and only relax it when necessary.

src/main/java/com/google/sps/servlets/Schema.java
```diff
@@ -0,0 +1,14 @@
...
+
+@AutoValue
+abstract class Schema {
+  static Schema create(String columnName, String schemaType, String nullable) {
+    return new AutoValue_Schema(columnName, schemaType, nullable);
+  }
+
+  abstract String columnName();
+  abstract String schemaType();
+  abstract String nullable();
```
jiaxinz-g: This nullable hasn't been really used, is that right? I'd imagine when it is being used, what we really need is a boolean. So please create it as a boolean instead of a String.

src/main/java/com/google/sps/servlets/TablesFromDatabaseServlet.java
```diff
@@ -0,0 +1,59 @@
...
+    response.setContentType(constants.TEXT_TYPE);
+
+    if (request.getParameter(constants.DATABASE_PARAM) != null) {
+      selectedDatabase = request.getParameter(constants.DATABASE_PARAM);
+    }
+
+    if (selectedDatabase == null || selectedDatabase.equals(constants.EMPTY_STRING)) {
+      response.sendRedirect(constants.NULL_REDIRECT);
+      return;
+    }
```
jiaxinz-g, line 41: What happens if the selectedDatabase is a string that represents a non-existing database? For normal user flows, this probably won't happen. But some edge cases could be users changing the URL, or there was a shared URL for a database that is not being supported anymore. We probably want to redirect and return early as well. 

Okay with updating this in a followup PR, please add a comment to track it. For example:

 //(TODO(issues/8)): check if the selectedDatabase is supported.

src/main/java/com/google/sps/servlets/DataFromDatabaseServlet.java
```diff
@@ -0,0 +1,166 @@
...
+  }
+
+  private void addDataToRowObject(String dataType, Row rowObject, String columnName, ResultSet resultSet) {
+    System.out.println(dataType);
+    switch (dataType) {
+      case "STRING":
+        rowObject.addData(columnName, resultSet.getString(columnName));
+        break;
+      case "BOOL":
+        rowObject.addData(columnName, constants.EMPTY_STRING + resultSet.getBoolean(columnName));
```
jiaxinz-g, line 130: missing a break here.

jiaxinz-g (commented):
Looks pretty good! For the comments that you plan to resolve in followup PRs, please add a TODO comment with the issue id to track it, and reply in the conversation that you will do it later.

gagomez (approved):

src/main/java/com/google/sps/servlets/Constants.java
sasymer: Done

src/main/java/com/google/sps/servlets/DataFromDatabaseServlet.java
sasymer: Done.

src/main/java/com/google/sps/servlets/Schema.java
sasymer: done

src/main/java/com/google/sps/servlets/TablesFromDatabaseServlet.java
sasymer: Ok, added this TODO for later PR

src/main/java/com/google/sps/servlets/Table.java
sasymer: Doing this in later PR.

src/main/java/com/google/sps/servlets/Table.java
sasymer: Done in later PR.

Issue #10, opened 2020-07-06
Authentication and Authorization 


jiaxinz-g:
Integrate with Google Authentication, verify the user is logged in and has access to the tool.

Issue #9, opened 2020-07-06
Splash Page


jiaxinz-g:
Create an input box for justification of using, handle the transition to the query result page, including passing the state and load the page.

PR #23, opened 2020-07-09
Audit commit

wdcroslen:
-Add text area for reason for use
-Add logic for retrieving databases and table selection in JS


src/main/java/com/google/sps/servlets/LoginServlet.java
```diff
@@ -0,0 +1,48 @@
...
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    response.setContentType("text/html");
+
+    UserService userService = UserServiceFactory.getUserService();
+
+    if (!userService.isUserLoggedIn()) {
+      String urlToRedirectToAfterUserLogsIn = "/login";
+      String loginUrl = userService.createLoginURL(urlToRedirectToAfterUserLogsIn);
+      response.getWriter().println("<p>Hello stranger.</p>");
+      response.getWriter().println("<p>Login <a href=\"" + loginUrl + "\">here</a>.</p>");
```
sasymer: Maybe should move HTML rendering info to a javascript file, but I am not sure. Maybe in this case it's okay because this looks similar to the Login servlet in one of the walkthroughs?

src/main/java/com/google/sps/servlets/ReasonServlet.java
```diff
@@ -0,0 +1,30 @@
...
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+/** Servlet that sends reason for use to /reason. (will change to write to database)*/
+@WebServlet("/reason")
+public final class ReasonServlet extends HttpServlet {
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    // Get the input from the form.
+    long timestamp = System.currentTimeMillis();
```
sasymer: On my website I had a line like this and G suggested I use Instant instead. https://docs.oracle.com/javase/tutorial/datetime/iso/instant.html

You can use Instant.now() to get a timestamp of the current time. If you call instant.toString() it gives a nice timestamp representation of the time.

src/main/webapp/script.js
```diff
@@ -22,3 +21,129 @@ function getDatabases(){
...
+    const search = window.location.search;
+    const queryString = tablesUrl + search;
+    var startIndex = 0;
+    var databaseString= "";
+    
+    for (var i = 0; i<search.length; i++){
+        if (search.charAt(i) == '='){
+            startIndex = i;
+        }
+        else if (search.charAt(i) == '&'){
```
sasymer: Can move the else if up a line

if {
} else if {
} else {
}

src/main/webapp/script.js
```diff
@@ -22,3 +21,129 @@ function getDatabases(){
...
+        else{
+             databaseString += search.charAt(i);
+        }
+    }
+
+    databaseString = databaseString.substring(startIndex);
+    fetch(queryString)
+    .then(response => response.json())
+    .then((list) => { 
+        const DbText = document.createElement('p');
```
sasymer: I think constants should start with a lowercase letter, but correct me if I'm wrong!

src/main/webapp/script.js
```diff
@@ -22,3 +21,129 @@ function getDatabases(){
...
+        else if (search.charAt(i) == '&'){
+            break;
+        }
+        else{
+             databaseString += search.charAt(i);
+        }
+    }
+
+    databaseString = databaseString.substring(startIndex);
+    fetch(queryString)
```
sasymer, line 47: This way definitely works! I'm not sure you even need to do a fetch here, since the database and tables are already in the URL of this page. As you got the database name from the search string, maybe you could also get the selected tables from there rather than doing a fetch?

sasymer (approved):
Overall looks good! Just a couple of suggestions, particularly about fetch request vs getting info from URL string.

src/main/java/com/google/sps/servlets/ReasonServlet.java
wdcroslen: Ok thank you! This seems very helpful!

src/main/webapp/script.js
wdcroslen: Done!

src/main/webapp/script.js
wdcroslen: Can't believe I missed this, fixed!

src/main/webapp/script.js
wdcroslen: That might improve running time so that is a good idea!

src/main/java/com/google/sps/servlets/LoginServlet.java
wdcroslen: It might be a good idea to render this in javascript to keep it cleaner

hilakey: Javascript is pretty good at creating DOM elements dynamically, which is how I did my filters! So I believe it will take off a lot of pressure off your servlet.

src/main/java/com/google/sps/servlets/LoginServlet.java
```diff
@@ -0,0 +1,48 @@
...
+import com.google.appengine.api.datastore.Query;
+import com.google.appengine.api.users.UserService;
+import com.google.appengine.api.users.UserServiceFactory;
+import java.io.IOException;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+@WebServlet("/login")
```
jiaxinz-g, line 19: This login page is a good starting point. It shows information of whether the user is logged in or not, and the information can be used in development and debugging. 

But eventually, Beetle will only have two pages as we designed, or maybe even better only one page if we can merge these two. So what happens is when user opens the first page, if the user is not logged in, the page is auto redirect to the login URL, and if user is logged in, we can display the user's account on the page. 

From a UX point of view, we would like the user need less clicks when using the tool. I am fine with merging this version of PR for now, and think about how to optimize the user experience when unifying pages, and integrating features this week.

src/main/java/com/google/sps/servlets/ReasonServlet.java
```diff
@@ -0,0 +1,34 @@
...
+import java.io.IOException;
+import java.time.Instant;
+import java.util.Arrays;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+/** Servlet that sends reason for use to /reason. (will change to write to database)*/
+@WebServlet("/reason")
```
jiaxinz-g, line 12: Similar comment as above, at the end, we should only have two pages. 

In this task https://github.com/googleinterns/step67-2020/issues/9, when user clicks submit button, this will direct user to the second page, and pass the reason to the second page. Since there is already a "/main-page" page in this project, consider using it and directing to this page with the reason coded in the parameter.

src/main/java/com/google/sps/servlets/ReasonServlet.java
```diff
@@ -0,0 +1,34 @@
...
+public final class ReasonServlet extends HttpServlet {
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    // Get the input from the form.
+    Instant timestamp = Instant.now();
+    String timeString = timestamp.toString();
```
jiaxinz-g, line 18: The timeString is not necessary, and we don't need to pass it in the main page.

jiaxinz-g (commented):
There are two features you are implementing in this single commit (https://github.com/googleinterns/step67-2020/issues/9 and https://github.com/googleinterns/step67-2020/issues/10), both started and not finished. I suggest you split the code into two PR for these two features. 

The first one is for the splash page, and transition to the main page. You will need to clean up the current content in the index page, and use the html that we designed for the splash page, consider migrating prototype login page here http://play-user-data-beetle.appspot.com/prototypes/login-page/login.html. The other part of this task is handle the transition to the main page, which we'd like to see the reason to use text is accessible in the main page. 

The second task is integrating with authentication, which can be done in a separate PR, and the goal is to have minimum user clicks, so we need to integrate UserService in the splash page and main page. Let's chat over gvc to clear up any questions or confusions.

src/main/java/com/google/sps/servlets/LoginServlet.java
wdcroslen: I will work on making the authentication automatic for both pages and not using a button or creating another page with the login.

src/main/java/com/google/sps/servlets/ReasonServlet.java
wdcroslen: Here as well I will focus on the user journey and also I will separate these commits.

hilakey (approved):
Everything checks out good job!

hilakey:
Also in your PR make sure to deploy on a live server so we can see the functionality and add the link

PR #24, opened 2020-07-10
Revise Table data structures and use AutoValue. Replace string concatenation with StringBuilder

sasymer:
This PR solely addresses comments left in my last PR. It does not change the functionality much--rather, it's intended to make my code better based on suggestions.
- Remove Row class, replacing List<Row> in Table with a 2D List
- Use AutoValue and Builder to make Table 
- Throughout DataFromDatabaseServlet, replace string concatenation (with for loop) to StringBuilder


src/main/java/com/google/sps/servlets/Constants.java
```diff
@@ -1,20 +1,25 @@
 package com.google.sps.servlets;
 
 class Constants {
```
gagomez, line 3: @jiaxinz-g 

Alright, so I am normally pro-constants but I think we need to draw a line somewhere. Are we going to have a constant SPACE = " "? (sarcasm intended) Some of these also get weird with spaces, some have them at both ends, others at one end, some have none. 

I suggest creating methods/classes that generate queries and dropping any constant that is equal to a single "atom" (e.g. close bracket, comma, from, null, true) - these can just be hardcoded as they are unlikely to change and just add an extra layer of indirection.

Having said that, I could be convinced otherwise.

src/main/java/com/google/sps/servlets/DataFromDatabaseServlet.java
```diff
@@ -33,59 +35,67 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
   private Schema createSchema(ResultSet resultSet) {
     String columnName = resultSet.getString(0);
     String schemaType = resultSet.getString(1);
+
+    // Remove length in parens (i.e. STRING(MAX) --> STRING)
     int indexOfOpeningParen = schemaType.indexOf("(");
     if (indexOfOpeningParen >= 0) {
       schemaType = schemaType.substring(0, indexOfOpeningParen);
     }
     String nullableString = resultSet.getString(2);
```
gagomez: Suggest naming this something like 'isNullableColumn' as the name 'nullableString' made me think that the variable itself was nullable - which also made me freak out until I realized what this was.

src/main/java/com/google/sps/servlets/DataFromDatabaseServlet.java
```diff
@@ -33,59 +35,67 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
     boolean nullable = false;
-    if (nullableString.toLowerCase().equals("true")) {
+    if (nullableString.toLowerCase().equals(constants.TRUE)) {
```
gagomez: According to this: https://cloud.google.com/spanner/docs/information-schema#information_schemacolumns
this column is set to either "YES" or "NO" - which I honestly think is ridiculous as it should have just been a boolean column.

src/main/java/com/google/sps/servlets/DataFromDatabaseServlet.java
```diff
@@ -33,59 +35,67 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
       nullable = true;
     }
     return Schema.create(columnName, schemaType, nullable);
   }
 
-  //TODO (issue 15): use StringBuilder instead
-  private String constructQueryString(List<Schema> schemas, Table tableObject, String table) {
-    String query = constants.SELECT;
+  private Statement constructQueryStatement(List<Schema> schemas, ImmutableList.Builder<String> columnNamesBuilder, String table) {
```
gagomez: Schema is a very general term (it could be a table's schema or even a database's schema) I would suggest naming this 'columnSchemas' which is what I think it is.

src/main/java/com/google/sps/servlets/Constants.java
```diff
@@ -1,20 +1,25 @@
...
+  static final String CLOSE_BRACKET = "]";
```
hilakey: Good way to make the code cleaner, especially with variables that are used consistently.

hilakey (approved):
Nice changes!

src/main/java/com/google/sps/servlets/DataFromDatabaseServlet.java
gagomez: (Somewhat long nit but also a great learning opportunity) This method actually does 2 things: 

1. Constructs query for table data
2. Creates a list of column names - by modifying the essentially output parameter `columnNamesBuilder`

For better and worse, Java does not support real output parameters which puts us in an unfortunate situation. We either mutate input parameters (as you do here), or we create a wrapper around all return types (Statement and ImmutableList<String>), or (my preferred solution) we split this into different functions. In this case, we can easily create the list of columnNames from the list of Schema - we may even replace the usage of columnNames with just using the Schema.

Mutating input parameters can be fairly dangerous as it is not always clear from the API. You can learn more about some of these dangers by researching the benefits of side-effect free functions.

src/main/java/com/google/sps/servlets/DataFromDatabaseServlet.java
```diff
@@ -95,71 +105,75 @@ private void initializeDatabase(String databaseName) {
...
+        rowBuilder.add(arrayToString);
         break;
       default:
-        rowObject.addData(columnName, "This type is not currently supported.");
+        rowBuilder.add(constants.UNSUPPORT_ERROR);
     }
   }
 
-  //TODO(issue 15): change to String Builder
   private String longArrayToString(long[] longArray) {
```
gagomez, line 181: There is an awesome method that may help you here - String.join() 

https://docs.oracle.com/javase/8/docs/api/java/lang/String.html#join-java.lang.CharSequence-java.lang.Iterable-

gagomez (approved):

src/main/java/com/google/sps/servlets/Constants.java
sasymer: That makes sense to me. It was starting to feel a bit extreme indeed. I was operating under the idea of not having "magic" values in code, but I also understand if you and @jiaxinz-g prefer hardcoding the "atom" values.

src/main/java/com/google/sps/servlets/DataFromDatabaseServlet.java

src/main/java/com/google/sps/servlets/DataFromDatabaseServlet.java
sasymer: Ah I remember reading that in the documentation but I guess forgot about it when I implemented this code. Fixed - done.

src/main/java/com/google/sps/servlets/DataFromDatabaseServlet.java
sasymer: That's true -- I should probably rename the object itself too.

sasymer: Thank you for this comment - that's really helpful to learn. I think it would make sense to get rid of the columnNames list altogether, since the names are already in ColumnSchema.

sasymer: For now I'm going to at least separate out the different methods. In a separate branch, I've already put the list of ColumnSchemas into the Table. So I'll get rid of the columnNames list in that branch for a new PR.

src/main/java/com/google/sps/servlets/DataFromDatabaseServlet.java
sasymer: Done!

src/main/java/com/google/sps/servlets/Constants.java
sasymer: Ok I took out the atom values.

PR #25, opened 2020-07-13
Create Filtering Panel

### LINKED ISSUE #11: Create the filtering panel
Display the identification columns (either user_id, device_id or both) based on the selected tables. Display filtering boxes for them, and require the user to input at least one box. On the panel, provide a way to switch between different filtering modes.

hilakey:
+ Create filter panel
+ Display filtering boxes for the columns of the selected tables
+ Add alert that prevents form submission if both user_id/device_id box is left empty
+ Create servlet that gets the selected table's columns

**UPDATE 6/15:**  Query and database are no longer hard coded and are retrieved from the URL parameters.
+ The form submits the information to the /main-page.html  url but it doesn't have a way to store the selected tables and which columns they belong to..

+ Also I didn't add a toggle to switch between filtering modes because I wanted to make sure the quick start filtering panel would be complete before implementing this feature.

**UPDATE 6/15:**  updated link --> [] https://play-user-data-beetle.uc.r.appspot.com/

Screenshot 2020-07-13 at 2 32 42 AM
Screenshot 2020-07-13 at 2 18 42 AM
Screenshot 2020-07-13 at 12 59 02 AM
Screenshot 2020-07-12 at 10 02 35 PM


src/main/java/com/google/sps/servlets/ColumnsFromTables.java
```diff
@@ -0,0 +1,83 @@
...
+/** Servlet that returns an HTML list of all the columns of based on the selected tables. */
+@WebServlet("/filters")
+public class ColumnsFromTables extends HttpServlet {
+    private static final String  TABLE_SELECT_PARAM =  "table-select";
+    private static final String DATABASE_SELECT_PARAM = "list-databases";
+    private static final String COMMA = ", ";
+    private static String SELECTED_TABLES = "";
+    private static String GET_COLUMNS_FROM_TABLES = "SELECT table_name, ARRAY_AGG(column_name) FROM information_schema.columns WHERE table_name in(";
+    private static String SELECTED_DATABASE = "";
+
```
sasymer: Extra space

src/main/java/com/google/sps/servlets/ColumnsFromTables.java
```diff
@@ -0,0 +1,83 @@
...
+/** Servlet that returns an HTML list of all the columns of based on the selected tables. */
+@WebServlet("/filters")
+public class ColumnsFromTables extends HttpServlet {
+    private static final String  TABLE_SELECT_PARAM =  "table-select";
+    private static final String DATABASE_SELECT_PARAM = "list-databases";
+    private static final String COMMA = ", ";
+    private static String SELECTED_TABLES = "";
+    private static String GET_COLUMNS_FROM_TABLES = "SELECT table_name, ARRAY_AGG(column_name) FROM information_schema.columns WHERE table_name in(";
+    private static String SELECTED_DATABASE = "";
```
sasymer: Once my recent PR gets merged, we could move these constants into the Constants.java file

src/main/java/com/google/sps/servlets/ColumnsFromTables.java
```diff
@@ -0,0 +1,83 @@
...
+    @Override
+    public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+      response.setContentType("application/JSON;"); 
+
+      Multimap<String, Object> data = ArrayListMultimap.create();
+
+      //Parsing the URL for the query parameters
+      String[] listOfTables = request.getParameterValues(TABLE_SELECT_PARAM);
+
+      String qry = "SELECT table_name, ARRAY_AGG(column_name) FROM information_schema.columns WHERE table_name in ('Singers', 'Songs', 'Albums') group by table_name";
```
sasymer: Could pull this string out into a constant

sasymer: Also, should eventually be able to get table names from the request. You could potentially get the selected tables from the URL and pass them into the fetch request.

src/main/webapp/MillenniaDev_script.js
```diff
@@ -0,0 +1,122 @@
...
+
+//Checks whether the user has left both of the user_id/device_id input box empty. If true, the user cannot submit the filter
+// query,if false submit
+function isFilterInputEmpty(){
+    var userID = document.getElementById("user_id").value;
+    var deviceID = document.getElementById("device_id").value;
+
+    //Return a message saying that the filters cannot be applied unless at least one of the inputs is filled.
+    if(userID == "" && deviceID == ""){
+        alert("Applying filters failed: Please input either user_id, device_id or both.");
```
wdcroslen: I think this message and functionality are really helpful! I'm wondering if there is ever a time where these won't be known and rather than restricting the filter we ask if they still want to continue?

src/main/webapp/MillenniaDev_script.js
```diff
@@ -0,0 +1,122 @@
...
+
+        checkboxes.style.border = "1px solid";
+        checkboxes.style.display = "none";
+        checkboxes.style.width = '200px';
+
+        var boxDiv = document.createElement('div');
+        boxDiv.style.padding ='4px';
+        checkboxes.appendChild(boxDiv);
+
+
```
wdcroslen: Maybe delete some of these extra new lines

src/main/webapp/MillenniaDev_script.js
```diff
@@ -0,0 +1,122 @@
...
+      if(list[i] != null){
+        dropdown.add(option);
+      } 
+    }
+  });
+}
+
+//Checks whether the user has left both of the user_id/device_id input box empty. If true, the user cannot submit the filter
+// query,if false submit
+function isFilterInputEmpty(){
```
wdcroslen: Since most of the new JS pertains to filters maybe we can make a new filter.js

src/main/webapp/main-page.html
```diff
@@ -3,13 +3,37 @@
   <head>
     <meta charset="UTF-8">
     <title>Main Page</title>
+<<<<<<< HEAD
+    <link rel="stylesheet" href="style.css">
+=======
+>>>>>>> 746cdf1ef08b56098ec984c4ccd80a5e865b063d
```
wdcroslen: I think you accidentally pasted your git token in the code here. (It happened to me once too)

wdcroslen (approved):
These changes look great! I really like the checkboxes and I am excited about future changes!

src/main/java/com/google/sps/servlets/ColumnsFromTables.java
```diff
@@ -0,0 +1,83 @@
...
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+/** Servlet that returns an HTML list of all the columns of based on the selected tables. */
+@WebServlet("/filters")
```
jiaxinz-g: rename this to be consistent with the purpose of the servlet.

src/main/java/com/google/sps/servlets/ColumnsFromTables.java
```diff
@@ -0,0 +1,83 @@
...
+public class ColumnsFromTables extends HttpServlet {
```
jiaxinz-g: Rename the class name to be ColumnsFromTablesServlet to reflect that this is a http servlet.

src/main/webapp/MillenniaDev_script.js
```diff
@@ -0,0 +1,122 @@
+//Returns a list of databases from Cloud Spanner as html dropdown
+function getDatabases(){
+  fetch("/databases").then(response => response.json()).then((list) => {
+    let dropdown = document.getElementById('list-databases');
+    dropdown.options.remove(0);
```
sasymer: What does this line do and why?

src/main/webapp/MillenniaDev_script.js
```diff
@@ -0,0 +1,122 @@
...
+    let defaultOption = document.createElement('option');
+    defaultOption.text = 'Select a Database';
+
+    dropdown.add(defaultOption);
+    dropdown.selectedIndex = 0;
+
+    let option;
+    
+    for (let i in list) {
```
sasymer: Suggestion - even though it's pretty clear what i does here, maybe call it index or something else to be even more clear.

src/main/webapp/MillenniaDev_script.js
sasymer: Nice idea!

src/main/webapp/MillenniaDev_script.js
```diff
@@ -0,0 +1,122 @@
...
+    //Return a message saying that the filters cannot be applied unless at least one of the inputs is filled.
+    if(userID == "" && deviceID == ""){
+        alert("Applying filters failed: Please input either user_id, device_id or both.");
+        return false;
+    }
+    return true;
+}
+
+function showFiltersPanel(){
+  var x = document.getElementById("filter-box");
```
sasymer: Again maybe make the variable name more clear

src/main/webapp/MillenniaDev_script.js
```diff
@@ -0,0 +1,122 @@
...
+  if (x.style.display === "none") {
+    x.style.display = "block";
+  } else {
+    x.style.display = "none";
+  }
+}
+
+//Returns checkbox dropdowns of the selected table's columns which can then be filtered
+function filterColumns() {
```
sasymer: This function is really long -- try to split it up into smaller functions if possible! Each smaller function could have its own specific purpose.

src/main/webapp/MillenniaDev_script.js
```diff
@@ -0,0 +1,122 @@
...
+    const tableFilters = document.getElementById('table-filters');
+    tableFilters.style.position = 'relative';
+    
+    //Create a select dropdown based on the table name as keys
+    for(var keys in tables){
+      var select = document.createElement('select');
+      select.style.width = '200px';
+      select.options.remove(0);
+      select.name = keys;
+      select.id = keys;
```
sasymer: For example maybe this could be its own function, because it sets up the select. Just an idea!

sasymer: Also eventually we could move styling to CSS rather than in Javascript :)

src/main/webapp/MillenniaDev_script.js
```diff
@@ -0,0 +1,122 @@
...
+        checkboxes.style.display = "none";
+        checkboxes.style.width = '200px';
+
+        var boxDiv = document.createElement('div');
+        boxDiv.style.padding ='4px';
+        checkboxes.appendChild(boxDiv);
+
+        
+
+        console.log(keys, col, tables[keys][0][col]);
```
sasymer: Remove the console log and the extra space

src/main/webapp/MillenniaDev_script.js
```diff
@@ -0,0 +1,122 @@
...
+      defaultOption.style.display = 'none'; //Hiding table name as a valid option to select
+
+      select.add(defaultOption);
+      select.selectedIndex = 0;
+
+            
+      let checkboxes = document.createElement('div');
+      
+      //Create checkboxes for each column of the table
+      for(var col in tables[keys][0]){
```
sasymer: This is another place you could maybe split into smaller functions. I think it could help code readers more easily see what's going on. You already kind of split this for loop into different sections, so perhaps think of how some of these could become their own functions (maybe one for creating the checkbox, one for creating and adding the label, etc)

src/main/webapp/MillenniaDev_script.js
```diff
@@ -0,0 +1,122 @@
...
+      }
+
+      //onclick event that will hide/show column filters
+      select.onclick = function() {
+        if (checkboxes.style.display === "none") {
+            checkboxes.style.display = "block";
+        } else {
+            checkboxes.style.display = "none";
+      }
+     };
```
sasymer: Indentation off a bit

sasymer (approved):
I know you put a lot of work into this, and the checkboxes are looking great! Looking forward to continuing work on filtering.

src/main/java/com/google/sps/servlets/ColumnsFromTables.java
```diff
@@ -0,0 +1,83 @@
...
+    private static final String  TABLE_SELECT_PARAM =  "table-select";
```
jiaxinz-g: Reuse and update the constants in https://github.com/googleinterns/step67-2020/blob/master/src/main/java/com/google/sps/servlets/Constants.java.

src/main/java/com/google/sps/servlets/ColumnsFromTables.java
jiaxinz-g: Use query as variable name instead of qry. query is not long at all, and a long variable name is even better than abbreviations which could be confusing. 

Also add a comment to track the unfinished feature, eg.
 //TODO(issue #25): removed hardcoded tables and use the ones from parameters.

src/main/webapp/MillenniaDev_script.js
hilakey: Done!

src/main/webapp/MillenniaDev_script.js
hilakey: Ah okay good idea

src/main/webapp/main-page.html
hilakey: Thanks for finding this, just fixed it.

src/main/webapp/MillenniaDev_script.js
hilakey: This clears out all the options of a dropdown so that you can manually input options as you see fit. By removing the option at (0) we're clearing space to add a default option like "Select columns" so that it can be modified to stay hidden or cannot be interacted with.

src/main/java/com/google/sps/servlets/ColumnsFromTables.java
hilakey: Done!

src/main/webapp/MillenniaDev_script.js
hilakey: Yes, I agree. I just modified it to reflect the change!

src/main/webapp/MillenniaDev_script.js
hilakey: I can try the reason all the styling and creating of elements is within one function is because they're dynamically created with the DOM, so they're not permanently on the html to be able to style them outside of js with a css file, because an error will come up with the elements as null. I'll do a bit more research into checking if it's possible to safely split this function up!

src/main/java/com/google/sps/servlets/ColumnsFromTables.java
```diff
@@ -0,0 +1,83 @@
...
+      Spanner spanner = SpannerOptions.newBuilder().build().getService();
+      DatabaseId db = DatabaseId.of("play-user-data-beetle", "test-instance", "example-db");
+      this.dbClient = spanner.getDatabaseClient(db);
+    }
+
+    @Override
+    public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+      response.setContentType("application/JSON;"); 
+
+      Multimap<String, Object> data = ArrayListMultimap.create();
```
jiaxinz-g: Declare the type specifically, e.g. List<String> instead of Object, so that it is clearer what type to expect from the variable. 

I mentioned Multimap because there are two ways you can do the query. You can either aggregate the columns per table in the query, like what you are doing now, and return just a Map. Or you can query "SELECT table_name, column_name .....", and parse the result in a Multimap. 

Here since you are aggregating the columns in the query, a regular hash map would be enough. For example:

Map<String, List<String>> data = new HashMap();

src/main/java/com/google/sps/servlets/ColumnsFromTables.java
```diff
@@ -0,0 +1,83 @@
...
+    private static String GET_COLUMNS_FROM_TABLES = "SELECT table_name, ARRAY_AGG(column_name) FROM information_schema.columns WHERE table_name in(";
+    private static String SELECTED_DATABASE = "";
+
+
+    DatabaseClient dbClient;
+
+    public void init() {
+      Spanner spanner = SpannerOptions.newBuilder().build().getService();
+      DatabaseId db = DatabaseId.of("play-user-data-beetle", "test-instance", "example-db");
+      this.dbClient = spanner.getDatabaseClient(db);
```
gagomez: Instead of hard coding the database, this should come from a URL param right?

src/main/java/com/google/sps/servlets/ColumnsFromTables.java
```diff
@@ -0,0 +1,83 @@
...
+      String[] listOfTables = request.getParameterValues(TABLE_SELECT_PARAM);
+
+      String qry = "SELECT table_name, ARRAY_AGG(column_name) FROM information_schema.columns WHERE table_name in ('Singers', 'Songs', 'Albums') group by table_name";
+
+      try (ResultSet resultSet =
+        dbClient
+            .singleUse() // Execute a single read or query against Cloud Spanner.
+            .executeQuery(Statement.of(qry))) {
+            while (resultSet.next()) {
+              data.put(resultSet.getString(0), resultSet.getStringList(1));
```
gagomez: I think we need more than just the column name right? The type and whether a column is a primary key should be important.

src/main/webapp/MillenniaDev_script.js
gagomez: +1

src/main/java/com/google/sps/servlets/ColumnsFromTables.java
hilakey: Done!

src/main/java/com/google/sps/servlets/ColumnsFromTables.java
hilakey: Done!

src/main/java/com/google/sps/servlets/ColumnsFromTables.java
hilakey: Just made these changes, thank you for pointing this out

src/main/webapp/MillenniaDev_script.js
hilakey: Done!

src/main/webapp/MillenniaDev_script.js
hilakey: Just fixed this

src/main/webapp/MillenniaDev_script.js
hilakey: Done!

src/main/java/com/google/sps/servlets/ColumnsFromTables.java
hilakey: Yes! In fact I didn't notice that we could use the URL param there, but I just modified this to show the change

PR #26, opened 2020-07-13
Toolbar

wdcroslen:
-Basic start for toolbar with buttons similar to prototype
-Data does not show until "Load Data" button is pressed
-copy link button for sharing query


src/main/webapp/main-page.html
```diff
@@ -5,11 +5,28 @@
...
-  <body onload="showDatabase()">
+  <body>
     <h1>Main Page</h1>
  
     <div id="data"></div>
+
+    <div id = "header">
+      <div class = "button-box">
+        <a href = "index.html"><button>Splash Page</button></a>
+        <button id = "filter-button" onmouseover = "luckyFilter()" onclick = "showFilterPanel()" onmouseout="unlucky()">Show Filters </button>
```
hilakey, line 23: This will be easy to merge with my code since it's just a button.

src/main/webapp/main-page.html
```diff
@@ -5,11 +5,28 @@
...
+        <button id = "data-button" onmouseover = "luckyData()" onclick = "showDatabase()" onmouseout="unlucky()">Load Data </button>
```
sasymer, line 24: Why are the functions called luckyData and unlucky?

src/main/webapp/script.js
```diff
@@ -103,3 +102,39 @@ function onLoad() {
...
+
+function luckyData() {
+    previousId="data-button";
+    previousText = document.getElementById("data-button").textContent;
+    document.getElementById("data-button").textContent = "I'm Feeling Lucky!";
+}
+
+function unlucky() {
+    document.getElementById(previousId).textContent = previousText;
+}
```
sasymer, line 218: Are these for the easter egg? If not maybe rename to fit the context more (I might just be confused)

sasymer (approved):
Toolbar looks good!

src/main/webapp/script.js
wdcroslen: Yeah these are more of an idea of the easter egg from what Guillermo mentioned about hovering over the button and the name changing! I should have clarify in a comment the reason for the bizarre naming.

jiaxinz-g (approved):

PR #27, opened 2020-07-14
Database Connector

sasymer:
- Use Singleton design pattern to create a DatabaseConnector (there is only one instance which is shared by various servlets)
- Initialize all databases' clients at once, and only once
- Reduces the expensive operation of initializing db clients


wdcroslen (approved):
The changes look good! Nice name changes and easy to follow!

jiaxinz-g (approved):

PR #29, opened 2020-07-15
Table column schema, fix merge

sasymer:
## Please review this AFTER my Database Connector PR, as it contains some commits from that PR

- replaces list of column names with list of ColumnSchema objects in Table
- fixes some merge duplication from last merged PR


src/main/java/com/google/sps/servlets/ColumnsFromTablesServlet.java
```diff
@@ -60,7 +60,7 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
           for(int i = 0; i < listOfTables.length; i++){
               if(i != listOfTables.length-1){
                 constant.SELECTED_TABLES = "\'" + listOfTables[i] + "\'";
```
jiaxinz-g: When using the constants strings from Constants.java, we don't need to create an instance of the class. Instead, we can just import the static variables in that class, and use it in servlets. 

import static com.google.sps.servlets.Constants. GET_COLUMNS_FROM_TABLES;

And if they are constants, it probably doesn't make sense to modify them, like you discovered today, "constant.QUERY = ... " could be problematic. Just declare a new String and construct the query would be fine.

src/main/java/com/google/sps/servlets/ColumnsFromTablesServlet.java
```diff
@@ -60,7 +60,7 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
-                constant.QUERY = constant.QUERY + constant.SELECTED_TABLES + constant.COMMA;
+                constant.QUERY = constant.QUERY + constant.SELECTED_TABLES + ", ";
```
jiaxinz-g: (A pending comment from the last PR, that I didn't get to send it out. I don't feel strong about this, if you keep the query this way for now, that is fine. )

Many of the atom constants can be avoided by formatting a string rather than concatenating strings. For example:

SCHEMA_INFO_SQL = "SELECT column_name, spanner_type, is_nullable FROM information_schema.columns WHERE table_name = '%s'"
And we can create the query with String.format(SCHEMA_INFO_SQL, table).

I think this makes it easier for readers to get what the queries are doing.

wdcroslen (approved):

src/main/java/com/google/sps/servlets/ColumnsFromTablesServlet.java
sasymer: Done

src/main/java/com/google/sps/servlets/ColumnsFromTablesServlet.java
sasymer: I like this idea a lot. I will make a note of this and try to implement it in my other servlets in a later PR.

src/main/java/com/google/sps/servlets/ColumnsFromTablesServlet.java
```diff
@@ -35,7 +35,6 @@
 @WebServlet("/columns-from-tables")
 public class ColumnsFromTablesServlet extends HttpServlet {
     Constants constant = new Constants();
-    
     DatabaseClient dbClient;
```
gagomez: Since this is not created in the constructor, let's have this as a regular variable in the function and pass it to the functions that need it.

src/main/java/com/google/sps/servlets/ColumnsFromTablesServlet.java
```diff
@@ -48,35 +47,35 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
-                constant.SELECTED_TABLES = "\'" + listOfTables[i] + "\'";
-                constant.QUERY = constant.QUERY + constant.SELECTED_TABLES + constant.COMMA;
-              }else{
-                  constant.SELECTED_TABLES = "\'" + listOfTables[i] + "\'";
-                  constant.QUERY = constant.QUERY + constant.SELECTED_TABLES;
-                }
-            }
-          constant.QUERY = constant.QUERY + constant.GROUP_BY_TABLE_NAMES;   
+      if (listOfTables.length == 1) {
+        query = query + constant.GET_COLUMNS_FROM_TABLES + "\'" + listOfTables[0] + "\'" + constant.GROUP_BY_TABLE_NAMES;
```
gagomez: Isn't this the same as the as what the else does for the case there is only one?

src/main/java/com/google/sps/servlets/ColumnsFromTablesServlet.java
```diff
@@ -48,35 +47,35 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
 
       try (ResultSet resultSet =
         dbClient
-            .singleUse() // Execute a single read or query against Cloud Spanner.
-            .executeQuery(Statement.of(constant.QUERY))) {
-            while (resultSet.next()) {
-              data.put(resultSet.getString(0), resultSet.getStringList(1));
-            }
+        .singleUse() // Execute a single read or query against Cloud Spanner.
+        .executeQuery(Statement.of(query))) {
```
gagomez: Tabs seem off here. If it is a continuation of the statement, tab with 4 spaces. Note the while below will still be tabbed with 2 spaces.

src/main/java/com/google/sps/servlets/ColumnsFromTablesServlet.java
```diff
@@ -48,35 +47,35 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
-                }
-            }
-          constant.QUERY = constant.QUERY + constant.GROUP_BY_TABLE_NAMES;   
+      if (listOfTables.length == 1) {
+        query = query + constant.GET_COLUMNS_FROM_TABLES + "\'" + listOfTables[0] + "\'" + constant.GROUP_BY_TABLE_NAMES;
+      } else {
+        query = query + constant.GET_COLUMNS_FROM_TABLES;
+        for (int i = 0; i < listOfTables.length; i++) {
+          if (i != listOfTables.length-1) {
+            selectedTables = "\'" + listOfTables[i] + "\'";
```
jiaxinz-g: Have you tried just put tables in quotes without backslash? I was trying to find cloud spanner java documents about taking the raw string, but didn't find any good one, reading from this example, it seems we can just pass in single quotes without the escape backslash. https://cloud.google.com/spanner/docs/getting-started/java#write-data.

src/main/java/com/google/sps/servlets/ColumnsFromTablesServlet.java
```diff
@@ -48,35 +47,35 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+            query = query + selectedTables + ", ";
+          } else {
+            selectedTables = "\'" + listOfTables[i] + "\'";
+            query = query + selectedTables;
+          }
         }
```
jiaxinz-g: Have you looked at  String.join()? https://docs.oracle.com/javase/8/docs/api/java/lang/String.html. This will be a good use for constructing selectedTables.

src/main/java/com/google/sps/servlets/DataFromDatabaseServlet.java
```diff
@@ -58,27 +58,19 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
     response.getWriter().println(json);
   }
 
-  private void checkTableHasColumns(List<ColumnSchema> columnSchemas) {
```
jiaxinz-g: You don't need to change this. Having the method taking in a List (more generic interface than ImmutableList) is the preferred way.

jiaxinz-g (approved):
There are better ways to construct the query string, but can be done in a followup PR.

src/main/java/com/google/sps/servlets/ColumnsFromTablesServlet.java
sasymer: I think this is mainly Millennia's servlet, so I'll pass this comment on to her and discuss together.

src/main/java/com/google/sps/servlets/ColumnsFromTablesServlet.java
sasymer: done

src/main/java/com/google/sps/servlets/ColumnsFromTablesServlet.java
sasymer: Yes - fixed

src/main/java/com/google/sps/servlets/ColumnsFromTablesServlet.java
sasymer: Done

src/main/java/com/google/sps/servlets/ColumnsFromTablesServlet.java
sasymer: @hilakey

src/main/java/com/google/sps/servlets/ColumnsFromTablesServlet.java
sasymer: @hilakey

src/main/java/com/google/sps/servlets/DataFromDatabaseServlet.java

PR #31, opened 2020-07-17
Constants, QueryFactory, Database selection (splash)

sasymer:
### A clean-up PR
- Move some specific (only used once) constants into their respective files/servlets
- Make QueryFactory class with static methods that build and return strings or statements for querying
- Also, fix Database selector (don't refresh the page after selecting)

### DIFF
--- a/src/main/java/com/google/sps/servlets/ColumnsFromTablesServlet.java
+++ b/src/main/java/com/google/sps/servlets/ColumnsFromTablesServlet.java
@@ -31,8 +31,6 @@ import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import static com.google.sps.servlets.Constants.DATABASE_PARAM;
-import static com.google.sps.servlets.Constants.GET_COLUMNS_FROM_TABLES;
-import static com.google.sps.servlets.Constants.GROUP_BY_TABLE_NAMES;
 import static com.google.sps.servlets.Constants.TABLE_SELECT_PARAM;
 
 /** Servlet that returns an HTML list of all the columns of based on the selected tables. */
@@ -51,21 +49,7 @@ public class ColumnsFromTablesServlet extends HttpServlet {
 
       DatabaseClient dbClient = DatabaseConnector.getInstance().getDbClient(database);
 
-      String query = "";
-      String selectedTables = "";
-
-      //TODO: use StringBuilder rather than string concatenation
-
-      query = query + GET_COLUMNS_FROM_TABLES;
-      for (int i = 0; i < listOfTables.length; i++) {
-        //TODO: check if backslash is actually needed here
-        selectedTables = "\'" + listOfTables[i] + "\'";
-        query = query + selectedTables;
-        if (i != listOfTables.length-1) {
-          query = query + ", ";
-        } 
-      }
-      query = query + GROUP_BY_TABLE_NAMES;   
+      String query = QueryFactory.getInstance().buildColumnsQuery(listOfTables);
 
       try (ResultSet resultSet =
           dbClient
--- a/src/main/java/com/google/sps/servlets/ColumnsFromTablesServlet.java
+++ b/src/main/java/com/google/sps/servlets/ColumnsFromTablesServlet.java
@@ -49,12 +49,12 @@ public class ColumnsFromTablesServlet extends HttpServlet {
 
       DatabaseClient dbClient = DatabaseConnector.getInstance().getDbClient(database);
 
-      String query = QueryFactory.getInstance().buildColumnsQuery(listOfTables);
+      Statement query = QueryFactory.getInstance().buildColumnsQuery(listOfTables);
 
       try (ResultSet resultSet =
           dbClient
           .singleUse() 
-          .executeQuery(Statement.of(query))) {
+          .executeQuery(query)) {
         while (resultSet.next()) {
           data.put(resultSet.getString(0), resultSet.getStringList(1));
         }
--- a/src/main/java/com/google/sps/servlets/DataFromDatabaseServlet.java
+++ b/src/main/java/com/google/sps/servlets/DataFromDatabaseServlet.java
@@ -44,10 +44,10 @@ public class DataFromDatabaseServlet extends HttpServlet {
     List<Table> tables = new ArrayList<>();
 
     for (String table : selectedTables) {
-      String columnQuery = QueryFactory.getInstance().buildSchemaQuery(table);
+      Statement columnQuery = QueryFactory.getInstance().buildSchemaQuery(table);
  
       try (ResultSet resultSet =
-        dbClient.singleUse().executeQuery(Statement.of(columnQuery))) {
+        dbClient.singleUse().executeQuery(columnQuery)) {
         ImmutableList<ColumnSchema> columnSchemas = initColumnSchemas(resultSet);
         
         Table.Builder tableBuilder = Table.builder().setName(table);
--- a/src/main/java/com/google/sps/servlets/DataFromDatabaseServlet.java
+++ b/src/main/java/com/google/sps/servlets/DataFromDatabaseServlet.java
@@ -53,6 +53,8 @@ public class DataFromDatabaseServlet extends HttpServlet {
         Table.Builder tableBuilder = Table.builder().setName(table);
         tableBuilder.setColumnSchemas(columnSchemas);
         Statement queryStatement = QueryFactory.getInstance().constructQueryStatement(columnSchemas, table);
+        System.out.println(queryStatement.toString());
+        
         executeTableQuery(tableBuilder, queryStatement, columnSchemas);
         
         Table tableObject = tableBuilder.build();
--- a/src/main/java/com/google/sps/servlets/DataFromDatabaseServlet.java
+++ b/src/main/java/com/google/sps/servlets/DataFromDatabaseServlet.java
@@ -53,7 +53,6 @@ public class DataFromDatabaseServlet extends HttpServlet {
         Table.Builder tableBuilder = Table.builder().setName(table);
         tableBuilder.setColumnSchemas(columnSchemas);
         Statement queryStatement = QueryFactory.getInstance().constructQueryStatement(columnSchemas, table);
-        System.out.println(queryStatement.toString());
         
         executeTableQuery(tableBuilder, queryStatement, columnSchemas);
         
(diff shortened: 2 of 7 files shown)


src/main/java/com/google/sps/servlets/QueryFactory.java
```diff
@@ -0,0 +1,51 @@
...
+
+import com.google.cloud.spanner.Statement;
+import java.util.List;
+
+final class QueryFactory {
+  
+  private static final QueryFactory instance = new QueryFactory();
+  private static final String GET_COLUMNS_FROM_TABLES = "SELECT table_name, ARRAY_AGG(column_name) FROM information_schema.columns WHERE table_name in(";
+  private static final String GROUP_BY_TABLE_NAMES = ") group by table_name";
+  private static final String SCHEMA_INFO_SQL = "SELECT column_name, spanner_type, is_nullable FROM information_schema.columns WHERE table_name = '";
```
gagomez: Let's keep the maximum length of a line in Java to 100 characters.

https://google.github.io/styleguide/javaguide.html#s4.4-column-limit

@jiaxinz-g 

Note: if these constants are only used in one place and are reasonably not going to be used in another place, I suggest not extracting them into constants. For example, it's unlikely to be another use of SCHEMA_INFO_SQL when the method buildSchemaQuery exists.

Another minor note, instead of string concatenation you could use String.format and pass in the table_name.

Alright last one: instead of concatenating the table name, let's bind parameters otherwise we may be vulnerable to SQL injection. See this for more details:
https://googleapis.dev/java/google-cloud-spanner/latest/index.html

src/main/java/com/google/sps/servlets/QueryFactory.java
```diff
@@ -0,0 +1,51 @@
...
+
+  private QueryFactory() {
+    // Private because static class
+  }
+
+  static QueryFactory getInstance() {
+    return instance;
+  }
+
+  static String buildSchemaQuery(String table) {
```
gagomez: Sometimes we return Strings, other times Statements. Since we are going to need to bind variables to the queries, it'll be a good idea to consolidate these into returning Statement rather than String.

gagomez (changes requested):

src/main/java/com/google/sps/servlets/QueryFactory.java
sasymer: Done - made them all into Statements

src/main/java/com/google/sps/servlets/QueryFactory.java
sasymer: I implemented all of these suggestions except for the last one. I tried binding but got an error. I had a sql String ".... where table-name = '@name'" and then used bind("name").to(tableName).

src/main/java/com/google/sps/servlets/QueryFactory.java
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  static QueryFactory getInstance() {
+    return instance;
+  }
+
+  //TODO: try to do binding w/ parameters
+  static Statement buildSchemaQuery(String table) {
+    StringBuilder query = new StringBuilder("SELECT column_name, spanner_type, is_nullable ");
+    query.append("FROM information_schema.columns WHERE table_name = '%s'");
```
jiaxinz-g: Using StringBuilder would be useful if we have some if conditions or for loops, but since here we always build the query string, and know what it looks like, consider build it directly. 

String query = "SELECT column_name, spanner_type, is_nullable FROM information_schema.columns WHERE table_name = '%s' "

src/main/java/com/google/sps/servlets/QueryFactory.java
```diff
@@ -0,0 +1,53 @@
...
+    return Statement.newBuilder(String.format(query.toString(), table)).build();
+  }
+
+  // Construct SQL statement of form SELECT <columns list> FROM <table>
+  static Statement constructQueryStatement(List<ColumnSchema> columnSchemas, String table) {
+    StringBuilder query = new StringBuilder("SELECT ");
+
+    for (ColumnSchema columnSchema : columnSchemas) {
```
jiaxinz-g: Try something like the following, please double check the syntax though. 

String query = String.format("Select @columns FROM %s ", table);
return Statement.newBuilder(query).bind("columns").to(String.join(", ", columnSchemas.stream().map(ColumnSchema::columnName).collect(Collectors.toList()))).build();

So here 
1. we have _query_ showing the whole query structure
2. try to leverage the statement bind with building the table name in the string
3. Using string join to build the list of columns in a string, but since the column name is in ColumnSchema class, use Java stream to unwrap and collect the names. Can have a local variable String columns = String.join(", ", columnSchemas.stream().map(ColumnSchema::columnName).collect(Collectors.toList()));

sasymer: I tried to do this, and the stream/map part seems to work: it prints out the correct column names. But then when I run the code I get an error saying Field not found: SingerId (which is the first column in the table).

src/main/java/com/google/sps/servlets/QueryFactory.java
sasymer: Done

src/main/java/com/google/sps/servlets/QueryFactory.java
sasymer: I learned that you can't bind table names.

src/main/java/com/google/sps/servlets/ColumnsFromTablesServlet.java
```diff
@@ -51,26 +49,12 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
-      for (int i = 0; i < listOfTables.length; i++) {
-        //TODO: check if backslash is actually needed here
-        selectedTables = "\'" + listOfTables[i] + "\'";
-        query = query + selectedTables;
-        if (i != listOfTables.length-1) {
-          query = query + ", ";
-        } 
-      }
-      query = query + GROUP_BY_TABLE_NAMES;   
+      Statement query = QueryFactory.getInstance().buildColumnsQuery(listOfTables);
```
hilakey, line 52: very neat statement!

hilakey (approved):
good changes!

src/main/java/com/google/sps/servlets/ColumnsFromTablesServlet.java
wdcroslen: ^^ Definitely this is a neat way to reduce code here

wdcroslen (approved):
Overall good changes!

src/main/java/com/google/sps/servlets/QueryFactory.java
sasymer: Done.

src/main/java/com/google/sps/servlets/ColumnsFromTablesServlet.java
sasymer: Thanks!

src/main/java/com/google/sps/servlets/DataFromDatabaseServlet.java
```diff
@@ -20,36 +20,41 @@
...
-        dbClient.singleUse().executeQuery(Statement.of(columnQuery))) {
-          
+        dbClient.singleUse().executeQuery(columnQuery)) {
         ImmutableList<ColumnSchema> columnSchemas = initColumnSchemas(resultSet);
         
         Table.Builder tableBuilder = Table.builder().setName(table);
         tableBuilder.setColumnSchemas(columnSchemas);
-        Statement queryStatement = constructQueryStatement(columnSchemas, table);
+        Statement queryStatement = QueryFactory.getInstance().constructQueryStatement(columnSchemas, table);
+        System.out.println(queryStatement.toString());
```
jiaxinz-g: nit: remove prints before merging.

src/main/java/com/google/sps/servlets/QueryFactory.java
```diff
@@ -0,0 +1,54 @@
...
+      query.append(columnSchema.columnName() + ", ");
+    }
+    query.deleteCharAt(query.length() - 1); //Get rid of extra space
```
gagomez: What about removing the ','?

src/main/java/com/google/sps/servlets/QueryFactory.java
```diff
@@ -0,0 +1,54 @@
...
+    getColumnsSql += "FROM information_schema.columns WHERE table_name in(";
+    StringBuilder queryBuilder = new StringBuilder(getColumnsSql);
+
+    for (int i = 0; i < listOfTables.length; i++) {
+      String selectedTables = "'" + listOfTables[i] + "'";
+      queryBuilder.append(selectedTables);
+      if (i != listOfTables.length-1) {
+        queryBuilder.append(", ");
+      } 
+    }
```
gagomez: This could be simplified with String.join

https://docs.oracle.com/javase/8/docs/api/java/lang/String.html#join-java.lang.CharSequence-java.lang.Iterable-

For example:
`String.join(", ", Arrays.stream(listOfTables).map(table -> String.format("'%s'", table)).collect(toList()));`

note: do not fit it all in one line

src/main/java/com/google/sps/servlets/QueryFactory.java
jiaxinz-g: Let's just use String formatting then. It would still be nice if you can avoid the for loop and removing the 
extra space manually with String join. 

String columns = String.join(", ", columnSchemas.stream().map(ColumnSchema::columnName).collect(Collectors.toList()));
String query = String.format("Select %s FROM %s ", columns, table);
return Statement.newBuilder(query).build();

jiaxinz-g (approved):

src/main/java/com/google/sps/servlets/QueryFactory.java
sasymer: Done

src/main/java/com/google/sps/servlets/DataFromDatabaseServlet.java

src/main/java/com/google/sps/servlets/QueryFactory.java
sasymer: Changed this to use string formatting instead

src/main/java/com/google/sps/servlets/QueryFactory.java
sasymer: Done

gagomez (approved):

PR #32, opened 2020-07-17
Rearrange Toolbar + Functionality

wdcroslen:
-Put login method after every fetch to ensure user remains logged in
-Added hide and show functions to toolbar buttons


src/main/webapp/database-tables.js
```diff
@@ -80,12 +80,37 @@ function createTable(name) {
...
+  showReason();
 }
 
 function login() {
   fetch("/login").then(response => response.json()).then((user) => {
     document.getElementById("user").innerText = user;
   });
+}
+
+//Note: this method will show the reason assuming the reason is the last thing in the querystring
```
sasymer: We should probably find a way to make sure it works no matter what. Here's a way to do this:

const params = new URLSearchParams(window.location.search);
String reason = params.get('reason');

We should probably also do a check to make sure that the params contains 'reason' - here's a link to more info about this: https://flaviocopes.com/urlsearchparams/

This way should also eliminate a lot of code and make it cleaner!

src/main/webapp/script.js
```diff
@@ -184,9 +184,17 @@ function createListElement(text) {
   return liElement;
 }
 
+var shareShowing = Boolean(false)
 function showShare() {
-    console.log("Show");
-    document.getElementById("share-form").classList.remove("invisible");
+    if (!shareShowing){
+        shareShowing = true;
```
sasymer: One small clean-up would be to do shareShowing = !shareShowing after the if-else statement instead of shareShowing = true and shareShowing = false.

src/main/webapp/main-page.html
```diff
@@ -17,10 +17,11 @@ <h1>Main Page</h1>
...
       <div class = "button-box">
-        <a href = "index.html"><button>Splash Page</button></a>
-        <button id = "filter-button" onmouseover = "luckyFilter()" onclick = "showFilterPanel()" onmouseout="unlucky()">Show Filters </button>
-        <button id = "data-button" onmouseover = "luckyData()" onclick = "showDatabase()" onmouseout="unlucky()">Load Data </button>
-        <button onclick = "showShare()" >Share Query</button>
+        <a href = "splash.html"><button>Back to Splash Page</button></a>
+        <button id = "filter-button" onClick="showFiltersPanel(); filterColumns()"> Show Filters </button>
+        <button id = "data-button" onmouseover = "luckyData()" onclick = "login(); showDatabase()" onmouseout="unlucky()">Load Data </button>
+        <button id = "share-button" onclick = "showShare()" >Share Query</button>
+        <button onclick = "clearFilters()"> Clear Filters </button>
```
sasymer: Tiny thing: some weird extra spaces here like "showShare()" >

sasymer (commented):
Nice changes. I left a comment on the showReason() function that might make the code cleaner and more flexible (in case the reason isn't actually at the end of the URL)

src/main/webapp/database-tables.js
wdcroslen: Done, thank you for the suggestion!

src/main/webapp/script.js
wdcroslen: Done.

src/main/webapp/main-page.html
wdcroslen: Fixed

src/main/webapp/database-tables.js
```diff
@@ -80,12 +80,55 @@ function createTable(name) {
...
+    var usersWithAccess = ["jiaxinz@google.com","gagomez@google.com",/*"test@example.com",*/"hilakey@google.com","sasymer@google.com","williamdc@google.com"];
+    console.log(currentUser[0]);
+    console.log("current user");
+    for (var i =0; i < usersWithAccess.length; i++){
+        if (currentUser[0] == usersWithAccess[i]){
+            return;
+        }
+    }
+    //TODO: Make sure this url is for the deployed version
+    window.location.assign("https://8080-25c0ac2a-87ce-4126-a0a0-8231fedddb09.us-central1.cloudshell.dev/denied.html");
```
gagomez: You could use window.location.hostname to find the right url. E.g. window.location.hostname + "/denied.html"

src/main/webapp/database-tables.js
```diff
@@ -80,12 +80,55 @@ function createTable(name) {
...
+    }
+
+    var usersWithAccess = ["jiaxinz@google.com","gagomez@google.com",/*"test@example.com",*/"hilakey@google.com","sasymer@google.com","williamdc@google.com"];
+    console.log(currentUser[0]);
+    console.log("current user");
+    for (var i =0; i < usersWithAccess.length; i++){
+        if (currentUser[0] == usersWithAccess[i]){
+            return;
+        }
+    }
```
gagomez: Let's move this check to the server and return whether the user has permission or not. Alternatively we can also redirect from the response to /denied.html and to the login URL in the case the user is not logged in.

src/main/webapp/main-page.html
```diff
@@ -17,10 +17,11 @@ <h1>Main Page</h1>
 
     <div id = "header">
       <div class = "button-box">
-        <a href = "index.html"><button>Splash Page</button></a>
-        <button id = "filter-button" onmouseover = "luckyFilter()" onclick = "showFilterPanel()" onmouseout="unlucky()">Show Filters </button>
-        <button id = "data-button" onmouseover = "luckyData()" onclick = "showDatabase()" onmouseout="unlucky()">Load Data </button>
-        <button onclick = "showShare()" >Share Query</button>
+        <a href = "splash.html"><button>Back to Splash Page</button></a>
```
gagomez: nit: no spaces around the =

src/main/webapp/main-page.html
```diff
@@ -30,23 +31,23 @@ <h4>Share Query</h4>
...
+        <h1> Filters</h1>
 
-          <input type="text" id="user_id" placeholder="Enter the user_id">
-          <div style="padding: 10px"></div> 
-          <input type="text" id="device_id" placeholder="Enter the device_id">
-          <div style="padding: 10px"></div>
-          <input type="search" placeholder="Search...">
-          <div style="padding: 10px"></div>
+        <input type="text" id="user_id" placeholder="Enter the user_id">
+        <div style="padding: 10px"></div>
```
gagomez: Didn't notice these before. What are they for?

src/main/webapp/script.js
```diff
@@ -224,5 +231,9 @@ function changeText() {
 function login(){
   fetch("/login").then(response => response.json()).then((user) => {
     document.getElementById("user").innerText = user;
+    var currentUser = user;
+    var myArray = ["jiaxinz@google.com","gagomez@google.com","hilakey@google.com","sasymer@google.com","williamdc@google.com"];
```
gagomez: This doesn't seem to be used.

src/main/webapp/splash.html
```diff
@@ -17,15 +17,18 @@ <h2>Beetle</h2>
 
       <div name="text_enter" class="text-enter">
         <div id="select-db">
-          <form action="/splash.html" method="GET" id="database-select-form" >
+          <form action="/splash.html" method="GET; login()" id="database-select-form" >
```
gagomez: I haven't seen this before. How does it work? Should we use onsubmit for this? Don't we want to execute login() onload and not necessarily when the form is submitted?

gagomez (dismissed):

src/main/webapp/database-tables.js
wdcroslen: This is very helpful, done!

src/main/webapp/main-page.html
wdcroslen: Done

src/main/webapp/splash.html
wdcroslen: The login onload still occurs. I believed that this does work similarly to an onsubmit. When the databases are selected I want login to be called again before the tables are shown. It might be better to use an onsubmit in case this does not work how I expected

src/main/webapp/script.js
wdcroslen: I noticed this method is only used outside of the prototype and it has been removed.

src/main/webapp/main-page.html
wdcroslen: Padding to space for the user/device id. These have been moved to the toolbar now

src/main/java/com/google/sps/servlets/LoginServlet.java
```diff
@@ -27,13 +27,13 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
     List<String> userEmail = new ArrayList<String>();
```
jiaxinz-g: What is the reason of having a list for userEmail? Why not just a single String?

Any anytime, there is only one user logged in, it is got from userService.getCurrentUser().getEmail(), and is what we would like to display on the website and check authorization against with. 

This likely caused the problem cause before user logged in, "Stranger" is always added into the list, and as the first one, in your javascript, you are only checking the first element in the list. 

I would also recommend some System.out.println in your local java code and check the user email string. Please remove the println before you commit.

wdcroslen: Initially I planned to include more information about the user but since the email is the only thing that is currently being stored I have changed this into a String. Instead of adding to a list the string now changes to either "Stranger" if the user is not logged in, the user's email, or "deny" if the user does not have access.

src/main/webapp/database-tables.js
wdcroslen: Most of this logic has been moved to the server but the redirects stayed in JS

src/main/java/com/google/sps/servlets/LoginServlet.java
```diff
@@ -9,39 +9,52 @@
...
-        String logoutUrl = userService.createLogoutURL(urlToRedirectToAfterUserLogsOut);
+      String loginUrl = userService.createLoginURL("/login");
+      response.getWriter().println("<p>Login <a href=\"" + loginUrl + "\">here</a>.</p>");
+      userEmail = "Stranger";
+    } else {
         String currentUserEmail = userService.getCurrentUser().getEmail();
-        userEmail.add(currentUserEmail);
+        userEmail = currentUserEmail;
+        if (!usersWithAccess.contains(currentUserEmail)){
+            userEmail="deny";
```
gagomez, line 52: Instead of using the email to deny access could we create an object that represent the user status? For example it can contain a few fields: a boolean isLoggedIn, a String email, a String logoutUrl, a String loginUrl, and a boolean hasAccess

src/main/webapp/filters.js
```diff
@@ -10,23 +10,38 @@ function isFilterInputEmpty(){
...
 }
 
-function showFiltersPanel(){
+function hideUserDeviceId() {
+    if (document.getElementById("user_id").style.display == "none") {
+         document.getElementById("user_id").style.display = "block";
+         document.getElementById("device_id").style.display = "block";
+    } else {
+    document.getElementById("user_id").style.display = "none";
+    document.getElementById("device_id").style.display = "none";
```
gagomez: indent these two in

jiaxinz-g (approved):

src/main/webapp/filters.js
sasymer: Indentation is off

src/main/java/com/google/sps/servlets/LoginServlet.java
```diff
@@ -9,39 +9,52 @@
...
+import java.util.Set;
+import java.util.HashSet;
 import javax.servlet.annotation.WebServlet;
 import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
 @WebServlet("/login")
 public class LoginServlet extends HttpServlet {
+    Set<String> usersWithAccess = new HashSet<String>();
```
sasymer, line 23: Nice change, adding this to the backend!

src/main/webapp/denied.html
```diff
@@ -0,0 +1,13 @@
...
+<html>
+  <head>
+    <meta charset="UTF-8">
+    <title>Access Denied</title>
+  </head>
+    
+  <body>
+    <h1>Access Denied!</h1>
+    <img src="sadPenguin.png" alt="penguin here">
+
```
sasymer: can delete this space

src/main/webapp/main-page.html
```diff
@@ -15,38 +15,37 @@ <h1>Main Page</h1>
...
+    <div id="alert" class="invisible">
+            <span class="closebtn" onclick="closeAlert()" >&times;</span> 
+            <strong>Link Copied!</strong>
+    </div>
+   
+    <div>
+        <p id="instruction">Enter userID or deviceID, then select filters.<p>
+        <input type="text" id="user_id" class="id" placeholder="Enter the user_id">
+        <input type="text" id="device_id" class="id" placeholder="Enter the device_id">
     </div>
```
sasymer: Looks like indentation is too far here

sasymer (approved):

src/main/webapp/denied.html
wdcroslen: Done

src/main/webapp/main-page.html
wdcroslen: Fixed

src/main/java/com/google/sps/servlets/LoginServlet.java
wdcroslen: Thank you!

src/main/webapp/filters.js
wdcroslen: Fixed!

src/main/webapp/database-tables.js
```diff
@@ -80,12 +80,28 @@ function createTable(name) {
...
   fetch("/login").then(response => response.json()).then((user) => {
     document.getElementById("user").innerText = user;
+    var currentUser = user;
+    var first = currentUser.split(" ");
+    if (currentUser == "deny"){
+        window.location.assign("/denied.html");
+    }
+    else if (currentUser == "Stranger") {
+        //This link first takes you to Google sign in and then continues back to splash page when signed in
+        window.location.assign("https://accounts.google.com/ServiceLogin?service=ah&passive=true&continue=https://uc.appengine.google.com/_ah/conflogin%3Fcontinue%3Dhttps://play-user-data-beetle.uc.r.appspot.com/splash.html");
```
gagomez, line 98: Isn't this the URL that we should get from the UserService?

PR #34, opened 2020-07-23
Sorting 

### LINKED ISSUE #15: Display the Raw Data
Show each queried table separately, and be able to sort by a column in the frontend. Display the query that was built and sent to the database.

sasymer:
- Completes front-end sorting by column
- Adds Table class to javascript for rendering and storing state
- Add SQL Strings representing query made to bottom of main page
- Please check table.js constructor to see if additional functions need to be bound. I read this page https://www.freecodecamp.org/news/this-is-why-we-need-to-bind-event-handlers-in-class-components-in-react-f7ea1a6f93eb/ for reference


jiaxinz-g (approved):

src/main/webapp/table.js
```diff
@@ -0,0 +1,149 @@
...
+//TODO: Check which additional functions need to be bound in constructor
+class Table {
+  constructor(dataTable, name, colSchemas, id) {
+    this.name = name;
+    this.colSchemas = colSchemas;
+    this.id = id;
+    this.dataTable = new Array(dataTable.length); // Number rows
+    this.sortDirections = new Array(colSchemas.length); // Number cols
+    this.makeTableWithTypes(dataTable);
+
```
wdcroslen: Maybe remove space here.

src/main/webapp/main-style.css
```diff
@@ -29,4 +29,17 @@
 
 .invisible {
   display: none;
+}
+
+#sql {
+  bottom: 0;
+  /*position: absolute;*/
```
wdcroslen: Remove comment, if it will be unused

wdcroslen (approved):
Really great changes! I like how you broke down the functionality into several easier to read methods.

src/main/webapp/database-tables.js
```diff
@@ -1,82 +1,46 @@
...
+  
     fetch(queryString)
     .then(response => response.json())
     .then((data) => { 
-      const dataArea = document.getElementById("data");
-    
+      document.getElementById("tables").innerText = '';
+      document.getElementById("sql").innerText = '';
+      
+      let id = 0;
```
hilakey, line 21: What does this id variable do?

src/main/webapp/main-page.html
```diff
@@ -44,9 +45,12 @@ <h1> Filters</h1>
...
           <div id="table-filters"></div>
 
-          <input type="submit" id="submit" value="submit" onClick=" return isFilterInputEmpty()">
+          <input type="submit" id="submit" value="submit" onClick="return isFilterInputEmpty()">
         </form>  
       </div>
     <div id="user"></div>
+    <div id="tables"></div>
+
+    <div id="sql"></div>
```
hilakey, line 52: So this is the div that displays what the user selected ?

src/main/webapp/table.js
```diff
@@ -0,0 +1,149 @@
...
+  setTable(dataTable) {
+    this.dataTable = dataTable;
+  }
+
+  remove() {
+    document.getElementById("table_" + this.name).remove();
+    document.getElementById("header_" + this.name).remove();
+  }
+
+  renderTable() {
```
hilakey: This function will be very important for our pagination and search bar functionality, so we should try to preserve this as much as possible

hilakey (approved):
Looks great!

src/main/webapp/database-tables.js
```diff
@@ -88,4 +52,37 @@ function login() {
...
+  table.flipSortDirection(index);
+  table.setTable(dataTable);
+
+  //need to empty and rerender
+  let tableIndex = 0;
+  for (tableIndex in tablesList) {
+    const tableObj = tablesList[tableIndex];
+    tableObj.remove();
+    tableObj.renderTable();
+  }
```
gagomez: Alternatively, we could add a "rerender()" method which removes the rows/headers of the table but keeps the div so that you don't have to render every table, just the one that changed.

gagomez: No need to do this right now though unless you think it's quick and easy.

gagomez: Note if we do this, we can have .setTable call .remove() and .rerender() though we have to be careful about when to call render() and when to call rerender().

Another alternative if for render() to check if the table div already exists and if it does, delete the children and readd them without deleting the div.

src/main/webapp/table.js
```diff
@@ -0,0 +1,149 @@
...
+          } else {
+            this.dataTable[rowIndex][col] = cell;
+          }
+        }
+      }
+    }
+  }
+
+  fetchTable() {
+    setTimeout(function() {
```
gagomez: what's the purpose of this setTimeout?

gagomez (approved):

src/main/webapp/database-tables.js
sasymer: It gives each table an id, makes it easier with sorting + rendering

src/main/webapp/main-page.html
sasymer: yes

src/main/webapp/table.js
sasymer: yep

src/main/webapp/database-tables.js
sasymer: Ok, I added a TODO and will try to do this later.

src/main/webapp/table.js
sasymer: I got rid of it, it wasn't actually needed

src/main/webapp/database-tables.js
sasymer: Actually, I already had the index of the table in question. So I can just call remove and render on it I think.

sasymer: Done

PR #35, opened 2020-07-23
Per table filtering 

### LINKED ISSUE #13: Support Per-Table Column Filtering
Display the tables users have selected, and display every column in the table with a filtering box. Apply user's input filtering to build the query when the user clicks on the button.

sasymer:
- Complete per table filtering for Strings, Ints, and Booleans (primitives)
- Add binding for WHERE condition values
- Add empty message to table if it has no rows with the applied filters
- Fix some indentation issues in javascript files

http://play-user-data-beetle.appspot.com/splash.html


src/main/java/com/google/sps/servlets/QueryFactory.java
```diff
@@ -24,14 +31,67 @@ static Statement buildSchemaQuery(String table) {
...
+    }
+    return loopCount;
   }
 
+  //TODO add types other than primitives
+  private static void appendCondition(String filterValue, Statement.Builder builder, String colType, String colName) {
+    String condString = colName + " = @" + colName;
+    switch (colType) {
+      case "STRING": 
+      case "DATE":
```
jiaxinz-g: I am curious does DATE typed column can be bind to a string value? It seems in https://googleapis.dev/java/google-cloud-spanner/latest/com/google/cloud/spanner/ValueBinder.html, there is a to(com.google.cloud.Date value) method, which hints that we need to create a Date variable first, like what you do for Timestamp.

src/main/java/com/google/sps/servlets/QueryFactory.java
```diff
@@ -24,14 +31,67 @@ static Statement buildSchemaQuery(String table) {
...
+      appendCondition(value, builder, colSchema.schemaType(), colName);
+      loopCount++;
+    }
+    return loopCount;
   }
 
+  //TODO add types other than primitives
+  private static void appendCondition(String filterValue, Statement.Builder builder, String colType, String colName) {
+    String condString = colName + " = @" + colName;
+    switch (colType) {
```
jiaxinz-g, line 74: Would be more comprehensive if the switch handles all types in https://googleapis.dev/java/google-cloud-spanner/latest/com/google/cloud/spanner/Type.Code.html, but I think it looks great already, can do this in a follow up PR if you have time.

jiaxinz-g (approved):

src/main/java/com/google/sps/servlets/QueryFactory.java
sasymer: It seemed to work when I treated it like a String.

sasymer: It seemed like it worked as a String when I tried it, but I converted to Date just to be safe.

src/main/java/com/google/sps/servlets/QueryFactory.java
sasymer: Ok! I now have all types working except Array, Bytes, and Struct. Wrote a TODO to add in a future PR.

wdcroslen (approved):
Great changes!

src/main/java/com/google/sps/servlets/DataFromDatabaseServlet.java
```diff
@@ -40,20 +40,32 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
-        ImmutableList<ColumnSchema> columnSchemas = initColumnSchemas(resultSet);
-        
+        ImmutableList<ColumnSchema> columnSchemas = initColumnSchemas(resultSet, selectedColsInTable);
+        Statement.Builder builder = Statement.newBuilder("");
+
         Table.Builder tableBuilder = Table.builder().setName(table);
         tableBuilder.setColumnSchemas(columnSchemas);
-        Statement queryStatement = QueryFactory.getInstance().constructQueryStatement(columnSchemas, table);
-        
+
```
hilakey, line 68: You could remove this extra space if needed.

hilakey (approved):
Nice work!

gagomez (approved):

PR #36, opened 2020-07-23
Write Audit to Database (Using Data-from-database Servlet)

wdcroslen:
-Delete Reason Servlet
-Use Data-from-database to write audit whenever data is loaded
-Audit Query column now SQL Statement


src/main/java/com/google/sps/servlets/ReasonServlet.java
```diff
@@ -1,27 +1,65 @@
...
   public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
-    // Get the input from the form.
+    response.setContentType(constants.TEXT_TYPE);
+    String reason = request.getParameter("reason");
+    String account = request.getParameter("account");
+    String query = request.getParameter("query");
+    queryString = request.getParameter("queryString");
+    selectedDatabase = "example-db";
+    this.dbClient = DatabaseConnector.getInstance().getDbClient(selectedDatabase);
+    insertUsingDml(this.dbClient,reason,account,query,response);
```
sasymer: I like your use of the helper method. I don't think you technically need to pass in dbClient since it's an instance variable and could be accessed by all methods in the class.

src/main/java/com/google/sps/servlets/ReasonServlet.java
```diff
@@ -1,27 +1,65 @@
...
-    String[] words = text.split("\\s*,\\s*");
-    response.setContentType("text/html;");
-    response.getWriter().println(Arrays.toString(words));
+    dbClient
+    .readWriteTransaction()
+    .run(
+        new TransactionCallable<Void>() {
+            @Override
+            public Void run(TransactionContext transaction) throws Exception {
+              String sql =
```
sasymer: Might be a good place to use String formatting. You could use String.format(" VALUES ('%s', '%s', '%s', '%s')", account, query, reason, timeString)
All the %s stand for a String, then you pass them into the String.format. It might make it feel a little cleaner.

sasymer (approved):
Nice changes! Your ReasonServlet code is really clean and easy to understand.

src/main/java/com/google/sps/servlets/ReasonServlet.java
wdcroslen: That's a great point I missed that, I fixed this now to have less parameters!

src/main/java/com/google/sps/servlets/ReasonServlet.java
wdcroslen: Thank you for the suggestion, this definitely looks a lot cleaner!

src/main/java/com/google/sps/servlets/ReasonServlet.java
```diff
@@ -1,27 +1,65 @@
...
 
-/** Servlet that sends reason for use to /reason. (will change to write to database)*/
+/** Servlet that writes Audit to database*/
 @WebServlet("/reason")
 public final class ReasonServlet extends HttpServlet {
+
+  DatabaseClient dbClient;
+  String queryString;
+  String selectedDatabase;
+  private Constants constants = new Constants();
```
jiaxinz-g: No need to create a Constants instance, just import static TEXT_TYPE and use it directly. 

import static com.google.sps.servlets.Constants.TEXT_TYPE;
response.setContentType(TEXT_TYPE);

src/main/java/com/google/sps/servlets/ReasonServlet.java
```diff
@@ -1,27 +1,65 @@
...
 import javax.servlet.http.HttpServletResponse;
 
-/** Servlet that sends reason for use to /reason. (will change to write to database)*/
+/** Servlet that writes Audit to database*/
 @WebServlet("/reason")
 public final class ReasonServlet extends HttpServlet {
+
+  DatabaseClient dbClient;
+  String queryString;
+  String selectedDatabase;
```
jiaxinz-g: No need for this class variable since we are always using "example-db", you can just delete it and when initiate dbClient, just do

this.dbClient = DatabaseConnector.getInstance().getDbClient("example-db");

src/main/java/com/google/sps/servlets/ReasonServlet.java
```diff
@@ -1,27 +1,65 @@
...
+    .readWriteTransaction()
+    .run(
+        new TransactionCallable<Void>() {
+            @Override
+            public Void run(TransactionContext transaction) throws Exception {
+              String sql =
+                  "INSERT INTO AuditLog (Account, Query, Reason, Timestamp) "
+                      + String.format(" VALUES ('%s', '%s', '%s', '%s')",account,query,reason,timeString);
+              long rowCount = transaction.executeUpdate(Statement.of(sql));
+              response.sendRedirect("/main-page.html" + queryString);
```
jiaxinz-g: I am curious if the sendRedirect is required here? What if you just delete this line?

src/main/java/com/google/sps/servlets/ReasonServlet.java
```diff
@@ -1,27 +1,65 @@
...
+    dbClient
+    .readWriteTransaction()
+    .run(
+        new TransactionCallable<Void>() {
+            @Override
+            public Void run(TransactionContext transaction) throws Exception {
+              String sql =
+                  "INSERT INTO AuditLog (Account, Query, Reason, Timestamp) "
+                      + String.format(" VALUES ('%s', '%s', '%s', '%s')",account,query,reason,timeString);
+              long rowCount = transaction.executeUpdate(Statement.of(sql));
```
jiaxinz-g: rowCount is not used below, so you can just do 

transaction.executeUpdate(Statement.of(sql));

src/main/java/com/google/sps/servlets/ReasonServlet.java
```diff
@@ -1,27 +1,65 @@
...
+              response.sendRedirect("/main-page.html" + queryString);
+              return null;
```
jiaxinz-g: No need to return null, just remove this line.

jiaxinz-g (commented):
A couple of coding stye comments, one bigger question is do we always need to send redirect in the response?

hilakey (approved):
Awesome! Good job :)

src/main/java/com/google/sps/servlets/ReasonServlet.java
wdcroslen: Done

src/main/java/com/google/sps/servlets/ReasonServlet.java
wdcroslen: Done

src/main/java/com/google/sps/servlets/ReasonServlet.java
wdcroslen: Done.

src/main/java/com/google/sps/servlets/ReasonServlet.java
wdcroslen: I have changed the code now and no longer need the redirect

src/main/java/com/google/sps/servlets/DataFromDatabaseServlet.java
```diff
@@ -35,6 +40,9 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
     response.setContentType(constants.TEXT_TYPE);
     selectedTables = request.getParameterValues(constants.TABLE_SELECT_PARAM);
     String databaseName = request.getParameter(constants.DATABASE_PARAM);
+    String reason = request.getParameter("reason");
```
sasymer, line 48: I'm not sure if we currently add the reason to the request. Did you make sure of this?

sasymer (approved):
Nice changes, and a nice short PR!

src/main/java/com/google/sps/servlets/DataFromDatabaseServlet.java
```diff
@@ -182,4 +194,23 @@ private String bytesToString(ByteArray bytes) {
...
+    .readWriteTransaction()
+    .run(
+        new TransactionCallable<Void>() {
+            @Override
+            public Void run(TransactionContext transaction) throws Exception {
+              String sql =
+                  "INSERT INTO AuditLog (Account, Query, Reason, Timestamp) "
+                      + String.format(" VALUES ('%s', '%s', '%s', '%s')",account,query,reason,timeString);
+              transaction.executeUpdate(Statement.of(sql));
+              return null;
```
jiaxinz-g, line 237: return null doesn't do anything here, you can just delete this line.

src/main/java/com/google/sps/servlets/DataFromDatabaseServlet.java
```diff
@@ -50,13 +58,17 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
         Table.Builder tableBuilder = Table.builder().setName(table);
         tableBuilder.setColumnSchemas(columnSchemas);
         Statement queryStatement = constructQueryStatement(columnSchemas, table);
+
+        queryAudit += queryStatement + "; ";
```
jiaxinz-g: Since this is inside a for loop, the number of String concatenation depends on the number tables queried, and it would be potentially large. Strings are immutable, use StringBuilder which is mutable in this case.

jiaxinz-g (approved):

src/main/java/com/google/sps/servlets/DataFromDatabaseServlet.java
wdcroslen: Yeah it is added!

gagomez (approved):

src/main/java/com/google/sps/servlets/DataFromDatabaseServlet.java
wdcroslen: Done!

PR #38, opened 2020-07-25
Demo from Today

sasymer:
Unfortunately this PR is really quite large. I tried to fix code formatting (basic things) as much as possible in a short time.
image


jiaxinz-g (approved):

src/main/java/com/google/sps/servlets/ColumnsFromTablesServlet.java
```diff
@@ -42,24 +44,34 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+      Statement query = QueryFactory.getInstance().buildFiltersQuery(listOfTables);
 
       try (ResultSet resultSet =
           dbClient
           .singleUse() 
           .executeQuery(query)) {
         while (resultSet.next()) {
+          for(int i = 0; i < resultSet.getStringList(1).size(); i++){
+              primaryKeyColumns.add(resultSet.getStringList(1).get(i));
+          }
```
gagomez: Can be simplified: `primaryKeyColumns.addAll(resultSet.getStringList(1));`

src/main/java/com/google/sps/servlets/ColumnsFromTablesServlet.java
```diff
@@ -42,24 +44,34 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+          data.put(resultSet.getString(0), resultSet.getStringList(2));
+                    //[table_name]        [columns_list]
```
gagomez: Comments like this can be avoided by using variables which won't grow out of date so easily. For example:

String tableName = resultSet.getString(0);
List<String> primaryKeys = resultSet.getStringList(1);
List<String> columns = resultSet.getStringList(2);

primaryKeyColumns.add(primaryKeys);
data.put(tableName, primaryKeys);
data.put(tableName, columns);

src/main/java/com/google/sps/servlets/QueryFactory.java
```diff
@@ -46,12 +45,20 @@ static void getWhereStatement(Statement.Builder builder, List<ColumnSchema> colu
...
     for (ColumnSchema colSchema : columnSchemas) {
       String colName = colSchema.columnName();
-      String filterValue = request.getParameter(table + "-" + colName);
-      loopCount = addWhere(filterValue, loopCount, builder, colName, colSchema);
+      if (colName.equals("UserId")) {
+        String idString = request.getParameter("user_id");
+        loopCount = addWhere(idString, loopCount, builder, colName, colSchema);
+      } else if (colName.equals("DeviceId")) {
+        String idString = request.getParameter("device_id");
+        loopCount = addWhere(idString, loopCount, builder, colName, colSchema);
```
gagomez, line 53: Would this work if the parameter for user_id and device_id were UserId and DeviceId, respectively?

src/main/java/com/google/sps/servlets/QueryFactory.java
```diff
@@ -126,4 +133,19 @@ static Statement buildColumnsQuery(String[] listOfTables) {
...
   }
-}
\ No newline at end of file
+
+  static Statement buildFiltersQuery(String[] listOfTables) {
+    String query = "SELECT table_name, ARRAY_AGG(column_name), ARRAY(SELECT column_name ";
+    query += "FROM information_schema.columns AS t ";
+    query += "WHERE r.table_name = t.table_name) as columns_list ";
+    query += "FROM information_schema.key_column_usage AS r ";
+    query += "WHERE table_name in (%s) group by table_name, constraint_name=\"PRIMARY_KEY\"";
```
gagomez, line 142: nit: (no need to change) this query will only return results for tables that *have* primary keys. We only have tables with primary keys in our database so it works but there is a way to restructure it so it works for tables without primary keys too.

src/main/java/com/google/sps/servlets/ReasonServlet.java
```diff
@@ -1,27 +1,63 @@
...
+import java.util.Enumeration;
+import java.util.List;
 import javax.servlet.annotation.WebServlet;
 import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
-/** Servlet that sends reason for use to /reason. (will change to write to database)*/
+/** Servlet that writes Audit to database*/
 @WebServlet("/reason")
```
gagomez, line 27: Just to be sure, this is changing to audit in the load data servlet right?

src/main/webapp/database-tables.js
```diff
@@ -7,8 +7,14 @@ function showDatabase() {
   if (!showing){
     showing = true;
     const search = window.location.search;
+    var searchParams = new URLSearchParams(search);
+    if (!(searchParams.has("user_id") && searchParams.has("device_id"))) {
```
gagomez: Does this mean that both user_id and device_id are required?

Also, we should show some information to the user to let them know why we did not do what they requests. E.g. "User ID and/or Device ID are required.

src/main/webapp/database-tables.js
```diff
@@ -66,13 +63,12 @@ function login() {
...
-    var first = currentUser.split(" ");
     if (currentUser == "deny"){
-        window.location.assign("/denied.html");
+      window.location.assign("/denied.html");
     }
     else if (currentUser == "Stranger") {
-        //This link first takes you to Google sign in and then continues back to splash page when signed in
-        window.location.assign("https://accounts.google.com/ServiceLogin?service=ah&passive=true&continue=https://uc.appengine.google.com/_ah/conflogin%3Fcontinue%3Dhttps://play-user-data-beetle.uc.r.appspot.com/splash.html");
+      //This link first takes you to Google sign in and then continues back to splash page when signed in
+      window.location.assign("https://accounts.google.com/ServiceLogin?service=ah&passive=true&continue=https://uc.appengine.google.com/_ah/conflogin%3Fcontinue%3Dhttps://play-user-data-beetle.uc.r.appspot.com/splash.html");
```
gagomez, line 66: Same as the comment on the previous PR, doesn't this URL come from the UserService? Shouldn't it be retrieved from the response?

src/main/webapp/filters.js
```diff
@@ -78,67 +46,253 @@ function filterColumns() {
...
+  primkeyDefaultOption.text = 'Primary Keys';
+  primkeyDefaultOption.style.display = 'none'; //Hiding table name as a valid option to select
+
+  primarykey_select.add(primkeyDefaultOption);
+  primarykey_select.selectedIndex = 0;
+
+  let primarykey_column_inputs = document.createElement('div');
+  primarykey_column_inputs.style.backgroundColor = 'white';
+ 
+  for(var keys in tables){
```
gagomez, line 99: What's the format of the data in tables? It seems really weird to me, it has a PrimaryKeys field and then a bunch for the other columns?

src/main/webapp/main-page.html
```diff
@@ -9,50 +9,69 @@
...
     <link rel="stylesheet" href="style.css">
+    <link rel="stylesheet" href="table-style.css">
+    <link href="https://fonts.googleapis.com/css2?family=Nunito:ital,wght@1,300&display=swap" rel="stylesheet">
+    <link href="https://fonts.googleapis.com/css2?family=Open+Sans&display=swap" rel="stylesheet">
   </head>
     
-  <body onload="mainLoad()">
-    <h1>Main Page</h1>
-
+  <body onload="mainLoad(); createFilters()">
```
gagomez: Can we just call createFilters in mainLoad?

gagomez (approved):

src/main/java/com/google/sps/servlets/ColumnsFromTablesServlet.java
sasymer: Done

src/main/java/com/google/sps/servlets/ColumnsFromTablesServlet.java
sasymer: Done

src/main/java/com/google/sps/servlets/QueryFactory.java
sasymer: Yes, would just have to change the parameter names in the javascript

src/main/java/com/google/sps/servlets/ReasonServlet.java
sasymer: Yes

src/main/webapp/database-tables.js
sasymer: @wdcroslen

src/main/webapp/database-tables.js
sasymer: I changed the logic to make sure that at least one is entered, and also added an alert.

src/main/webapp/main-page.html
sasymer: Yes - done

src/main/webapp/filters.js
sasymer: @hilakey

src/main/webapp/database-tables.js
wdcroslen: Initially I sent the redirect from the Servlet however when I deployed the server I got a "Redirect has been blocked by CORS policy: No Access Control Allow Origin" error message which is why it was moved to JS

PR #40, opened 2020-07-28
Data Conversion

### LINKED ISSUE #16: Data Conversion for Better Displaying
There are a couple things we would like to take care when displaying the data, for example:

1. read byte from the database, and then convert them to protobufs for displaying
2. read timestamp in long type, and convert them to human-readable timestamp type in PST 
3. read integer, and convert them to protobuf enums, and display human-readable enum values

wdcroslen:
Files Changes:
  database-tables.js      :     Lines 32 -47 ,  85- 145,
  table-style.css         : Line  15

- convert "proto"  (Json) to better viewable data
- read long timestamp, and convert to human-readable timestamp in PST
- read integer, and display human-readable enum values
Genre_EnumChanges
Millis ProtoChanges

### DIFF
--- a/src/main/webapp/database-tables.js
+++ b/src/main/webapp/database-tables.js
@@ -82,34 +82,34 @@ function login() {
 //TODO: Pass another paramater to combine data conversion functions 
 function dataConversionMillis(column,tableData){
     tableData.columnSchemas[column].schemaType = "TIMESTAMP";
-    for (var i =0; i<tableData.dataTable.length; i++) {
-      var time = parseInt(tableData.dataTable[i][column]);
+    for (var index=0; index<tableData.dataTable.length; index++) {
+      var time = parseInt(tableData.dataTable[index][column]);
       var date = new Date(time); 
       date.setHours(date.getHours() - 2); //Convert from default Central Time to PST
-      tableData.dataTable[i][column] = date.toISOString();
+      tableData.dataTable[index][column] = date.toISOString();
     } 
     return tableData;
 }
 
 function dataConversionGenreEnum(column,tableData){
     tableData.columnSchemas[column].schemaType = "STRING";
-    for (var i =0; i<tableData.dataTable.length; i++) {
-      var number = parseInt(tableData.dataTable[i][column]);
+    for (var index=0; index<tableData.dataTable.length; index++) {
+      var number = parseInt(tableData.dataTable[index][column]);
       switch(number) {
         case 1:
-            tableData.dataTable[i][column] = "Rock";
+            tableData.dataTable[index][column] = "Rock";
             break;
         case 2:
-            tableData.dataTable[i][column] = "Jazz";
+            tableData.dataTable[index][column] = "Jazz";
             break;
         case 3:
-            tableData.dataTable[i][column] = "Classical";
+            tableData.dataTable[index][column] = "Classical";
             break;
         case 4:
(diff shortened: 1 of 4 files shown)


gagomez (commented):
Looks like this includes all the changes in the large commit? Can you share what has changed?

src/main/webapp/database-tables.js
```diff
@@ -22,33 +28,34 @@ function showDatabase() {
...
-        }
 
         const colSchemas = tableData.columnSchemas;
+        var finalDataTable = tableData;
+    
+        for (var i =0; i<colSchemas.length; i++) {
+            if (colSchemas[i].columnName.endsWith('Millis')){
+                var finalDataTable = dataConversionMillis(i,finalDataTable);
+            } else if (colSchemas[i].columnName == "Genre") {
+                var finalDataTable = dataConversionEnum(i,finalDataTable);
```
jiaxinz-g: Rename dataConversionEnum to dataConversionGenreEnum. 

Enum conversion without know what the enum is hard to generalize, we will probably need a map from the column name to the enums. For now, I think this looks great! Just try to specify that this works for Genre and will need to extend to other enums.

jiaxinz-g (approved):

src/main/webapp/database-tables.js
wdcroslen: Thank you, this has been changed!

src/main/webapp/database-tables.js
```diff
@@ -33,15 +33,26 @@ function showDatabase() {
         const isEmpty = tableData.isEmpty;
 
         const colSchemas = tableData.columnSchemas;
+        var finalDataTable = tableData;
+    
+        for (var i =0; i<colSchemas.length; i++) {
```
sasymer, line 38: nit: could rename i to index

src/main/webapp/database-tables.js
```diff
@@ -68,6 +79,67 @@ function login() {
   });
 }
 
+//TODO: Pass another paramater to combine data conversion functions 
+function dataConversionMillis(column,tableData){
+    tableData.columnSchemas[column].schemaType = "TIMESTAMP";
+    for (var i =0; i<tableData.dataTable.length; i++) {
```
sasymer: Same comment (maybe index instead of i but not a big deal)

src/main/webapp/database-tables.js
```diff
@@ -68,6 +79,67 @@ function login() {
...
+function dataConversionProto(column,tableData) {
+    //TODO: work on clicking the { to open and close
+    for (var i = 0; i<tableData.dataTable.length; i++) {
+      var json = tableData.dataTable[i][column];
+      var newString = "";
+      var tab = Boolean(false);
+
+      for (var j = 0; j<json.length; j++) {
+          if (json[j] == ',' ) {
+              newString += (Boolean(tab) ? (json[j] + "\n"+"\t") : (json[j] + "\n"));
```
sasymer: Could you make the "\n" + "\t" into "\n\t"?

sasymer (approved):
Looks good!

src/main/webapp/database-tables.js
wdcroslen: Done!

src/main/webapp/database-tables.js
wdcroslen: Its more clear so I changed it!

PR #44, opened 2020-07-30
Id requirements

sasymer:
- Fix userid/deviceid requirements, depending on various cases of table columns
- Add check to backend, making sure the request has a non-empty and non-null userID or deviceID
- Add error message to screen if fetched data is empty (ie userID and deviceID are invalid)
Screenshot 2020-07-30 at 12 12 47 PM


src/main/java/com/google/sps/servlets/DataFromDatabaseServlet.java
```diff
@@ -42,7 +42,11 @@
 
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
-    //TODO: check to make sure there is userid or deviceid in request
+    if (request.getParameter("user_id") == null || request.getParameter("user_id") == "") {
```
jiaxinz-g: Consider extracting this into a helper method, isParameterEmpty(String parameterName), and inside it, you can check if the parameter is null or empty string. 

And here we can simply have  if(isParameterEmpty(user_id) && isParameterEmpty("device_id")).

jiaxinz-g (approved):

src/main/java/com/google/sps/servlets/DataFromDatabaseServlet.java

wdcroslen (approved):
Good changes, this PR addresses some of the demo issues so its great!

src/main/java/com/google/sps/servlets/DataFromDatabaseServlet.java
```diff
@@ -42,7 +42,11 @@
 
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
-    //TODO: check to make sure there is userid or deviceid in request
+    String userID = request.getParameter("user_id");
```
gagomez: Weren't we changing these to UserId and DeviceId to avoid the custom logic on the server?

gagomez (approved):

src/main/java/com/google/sps/servlets/DataFromDatabaseServlet.java
sasymer: Yes, done now

PR #45, opened 2020-07-30
Update pages, get rid of unused files

sasymer:
- Get rid of example files
- Make splash page into index.html


jiaxinz-g (approved):

wdcroslen (approved):

src/main/webapp/index.html
```diff
@@ -2,33 +2,31 @@
 <html>
   <head>
     <meta charset="UTF-8">
-    <title>Play User Data Beetle</title>
-    <script src="script.js"></script>
-    <link rel="stylesheet" href="style.css">
+    <title>Splash Page</title>
```
gagomez: This should remain Play User Data Beetle right?

gagomez (approved):

src/main/webapp/index.html
sasymer: done

